PR #4, opened 2015-07-21
Add support for delayed jobs if the Publisher supports the feature

jlogsdon:
SQS has a `DelaySeconds` attribute for individual messages allow you to immediately mark a message as invisible for that amount of time. We need this functionality for [TPAT work with Google](https://jira.tapjoy.net/browse/DMMEAS-17). We can also use this to exponentially (up to a limit) back off retries for the (eg) Downloads queue!

If a Publisher doesn't support a delay we're currently warning to the `Chore.logger` and ignoring the setting. In the case of SQS, which has a maximum delay of 900 seconds, we warn if the given delay is greater than the maximum and log a warning.

/to @StabbyCutyou 
/cc @jjrussell, @andyleclair, @nwmartin

### DIFF
--- a/lib/chore/job.rb
+++ b/lib/chore/job.rb
@@ -16,7 +16,7 @@ module Chore
       # Throw a RejectMessageException from your job to signal that the message should be rejected.
       # The semantics of +reject+ are queue implementation dependent.
     end
-    
+
     def self.job_classes #:nodoc:
       @classes || []
     end
@@ -42,9 +42,9 @@ module Chore
 
     module ClassMethods
       DEFAULT_OPTIONS = { }
-      
+
       # Pass a hash of options to queue_options the included class's use of Chore::Job
-      # +opts+ has just the one required option. 
+      # +opts+ has just the one required option.
       # * +:name+: which should map to the name of the queue this job should be published to.
       def queue_options(opts = {})
         @chore_options = (@chore_options || DEFAULT_OPTIONS).merge(opts_from_cli).merge(opts)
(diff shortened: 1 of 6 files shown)


lib/chore/job.rb
```diff
@@ -108,11 +115,22 @@ def perform(*args)
...
-      @chore_publisher.publish(self.class.prefixed_queue_name,self.class.job_hash(args))
-      self.class.run_hooks_for(:after_publish,*args)
+      push_to_publisher(args)
+    end
+
+    # Use the current configured publisher to send this job into a queue with the given delay period (in seconds). If
+    # the given publisher does not support delayed jobs a warning should be issued and the job will be queued without a
+    # delay.
+    def perform_delayed(delay, *args)
+      push_to_publisher(args, delay: delay)
```
StabbyCutyou: nitpick - don't mix hash syntaxes. We use rocket everywhere else in the file.

lib/chore/job.rb
```diff
@@ -108,11 +115,22 @@ def perform(*args)
...
     end
 
+    private
+    def push_to_publisher(args, options={})
+      self.class.run_hooks_for(:before_publish, *args)
+      @chore_publisher ||= self.class.options[:publisher]
+      @chore_publisher.publish(self.class.prefixed_queue_name, self.class.job_hash(args), options)
```
StabbyCutyou: This will break any existing chore publisher that uses the existing interface which is not in the chore-core gem (so, everything in chore-messaging). You'll want to update those as well.

jlogsdon: How? The `perform_async` method is still there and has the same signature... or do you mean the `publish` method? That is an optional argument (defaults to an empty hash).

lib/chore/job.rb
jlogsdon: Changed.

lib/chore/job.rb
jlogsdon: Oh wait, nevermind. Because chore-messaging adds *new* publishers. I'll add a sibling PR over there.

A minor version bump (1.x) will probably be good for this given the API change.

lib/chore/queues/sqs/publisher.rb
```diff
@@ -15,9 +15,27 @@ def initialize(opts={})
...
+          # SQS only supports a delay range of 0-900 seconds. Make sure we're in the range. We also only put it into the
+          # options Hash if its set to prevent any weird SQS behaviors.
+          # NOTE may be okay pushing the key with a `nil` value...
+          if delay_seconds
+            if delay_seconds < 0
+              delay_seconds = 0
+            elsif delay_seconds > 900
+              Chore.logger.warn("SQS does not support a delay longer than 15 minutes (900 seconds)! " <
+                                "Reducing the delay to the maximum supported.")
+              delay_seconds = 900
```
StabbyCutyou: I actually think we want to raise an error here. Any time we do a thing SQS doesn't support, we raise an error from what I recall (either we do, or SQS does because we're not guarding against it)

For example: https://github.com/Tapjoy/chore_internal/blob/master/lib/chore/queues/sqs/consumer.rb#L20

StabbyCutyou:
Oh, you've got this targeted against the wrong repo:

https://github.com/Tapjoy/chore_internal

Then we do a merge to this one.

lib/chore/queues/sqs/publisher.rb
jlogsdon: Hm, okay. I guess the FileSystem queue should also raise in this case, then?

jlogsdon:
Ugh. UGH!

GitHub: MOVE MY PR!

StabbyCutyou:
Yea :/ it's not great. We need to get that policy adjusted

PR #9, opened 2015-10-29
Fix mistake on hooks doc

tannerburson:
Example in the Hooks portion of the doc is wrong.

ping @StabbyCutyou


mattcarbone:
lgtm

PR #11, opened 2015-11-10
Only log to SQS at the same level as Chore

tannerburson:
This creates real problems when relying on STDOUT and piping for log handling. There's literally no way to stop AWS from spamming your log with messaging without something like this.

ping @StabbyCutyou

### DIFF
--- a/lib/chore/queues/sqs/consumer.rb
+++ b/lib/chore/queues/sqs/consumer.rb
@@ -118,7 +118,7 @@ module Chore
             :access_key_id => Chore.config.aws_access_key,
             :secret_access_key => Chore.config.aws_secret_key,
             :logger => Chore.logger,
-            :log_level => :debug)
+            :log_level => Chore.config.log_level)
         end
 
         def sqs_polling_amount
--- a/lib/chore/queues/sqs/consumer.rb
+++ b/lib/chore/queues/sqs/consumer.rb
@@ -114,11 +114,12 @@ module Chore
 
         # Access to the configured SQS connection object
         def sqs
-          @sqs ||= AWS::SQS.new(
+          sqs_options = {
             :access_key_id => Chore.config.aws_access_key,
-            :secret_access_key => Chore.config.aws_secret_key,
-            :logger => Chore.logger,
-            :log_level => Chore.config.log_level)
+            :secret_access_key => Chore.config.aws_secret_key
+          }
+          sqs_options.merge!(:logger => Chore.logger, :log_level => :debug) if Chore.config.log_level == Logger::DEBUG
+          @sqs ||= AWS::SQS.new(sqs_options)
         end
 
         def sqs_polling_amount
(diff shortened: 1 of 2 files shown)


StabbyCutyou:
Makes a ton of sense.

Working on correcting an issue with the internal/external nature of the Chore repo, which I'm trying to force an answer on today. I should be able to merge this in and do a release either way (and will get your other PRs in as well) within a day or two, considering the other work currently going on.

tannerburson:
I've got some changes to this, as the AWS gem is full of lies about how it logs. Stay tuned.

tannerburson:
This could be made slightly more efficient by memoizing the entire operation of the function instead of just the output. But I didn't quite go that far. Yet.

StabbyCutyou:
Cool

Are you happy with it as is, then?

tannerburson:
Yes, I'm actually using this in production right now.

lib/chore/queues/sqs/consumer.rb
```diff
@@ -114,11 +114,12 @@ def queue_timeout
...
         def sqs
-          @sqs ||= AWS::SQS.new(
+          sqs_options = {
             :access_key_id => Chore.config.aws_access_key,
-            :secret_access_key => Chore.config.aws_secret_key,
-            :logger => Chore.logger,
-            :log_level => :debug)
+            :secret_access_key => Chore.config.aws_secret_key
+          }
+          sqs_options.merge!(:logger => Chore.logger, :log_level => :debug) if Chore.config.log_level == Logger::DEBUG
```
StabbyCutyou: Would it not be best to keep having it use the Chore.logger, and simply default it's internal log level to match what chore is configured to do?

tannerburson: If it worked, yes. Near as I can tell despite having a `log_level` setting,
the AWS gem (or at least the SQS portion) either ignores it entirely, or
uses at least `info` level logging for everything. There is no difference
between `info` and `debug` in my testing.

StabbyCutyou:
Oh jeez, I totally forgot about this, sorry.

So long as it's been tested and is a known issue, it's fine with me. @theo-lanman or @obrie with a second opinion, otherwise I'm good with it.

obrie:
Works for me

tannerburson:
Eh, hold off a bit. While this works, it doesn't work good enough. This has
just shunted the logging off to somewhere else, which in some cases may be
better, and in others may be worse. I need to actually dig into their
logging code to see what the correct set of switches will be to make it
behave.

StabbyCutyou:
As someone who has read through the AWS SDK ruby code... hey, you have fun
with that.

tannerburson:
I'm closing this, as the solution presented works-ish, but not for cases anyone will actually care about. I've got a bigger WIP that covers this,  and about a thousand other things that I'll submit later.

PR #12, opened 2015-12-08
Final internal branch

StabbyCutyou:
ping @jlogsdon @nwmartin 

Can you guys confirm that these are your changes, and you're happy with them being merged into the public repo, which is about to become the only repo?


nwmartin:
Looks like my changes.

jlogsdon:
:+1: would love a public commit telling rspec to shut-up.

Issue #7, opened 2015-10-29
Docs don't mention how to manage Chore processes


tannerburson:
Docs do a good job of describing the signals that chore can handle, but there's no mention of how to find the PID of a running chore master. And, since Chore currently doesn't record a pidfile at all, a recommended Upstart or similar configuration would be useful.

Issue #10, opened 2015-10-29
Doesn't work out of the box in Rails


tannerburson:
Out of the box Chore won't work with a Rails app. Job classes in a Rails must be eager loaded in all environments for Chore to work. By default adding it to the eager load path won't work in development, as Rails lies and doesn't REALLY eager load in dev (to allow for class reloading mostly).

So, Chore should document how to setup eager load, and/or document the directory glob require one-liner that's been used in the past.

PR #14, opened 2015-12-09
Updating the docs to include notes on upstart, chore pids

StabbyCutyou:
This is to address #7 

There is probably more we could do here, but this gets something started that could be improved upon

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,11 +52,32 @@ Note that you can use one or the other but not both. Chore will quit and make fu
 
 ### Tips for configuring Chore
 
+For Rails, it can be necessary to add the directory you store your jobs in to the eager loading path,
+found in application.rb. You likely need a similar approach for most apps using jobs, unless you places
+them into a directory that is already eager loaded by default. One example of this might be:
+
+```ruby
+config.eager_load_paths += File.join(config.root, "app", "jobs")
+```
+
+However, due to the way eager_load_paths works in Rails, this may only solve the issue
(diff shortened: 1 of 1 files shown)


StabbyCutyou:
Added a note to address issue #10 as well

README.md
```diff
@@ -52,6 +52,14 @@ Note that you can use one or the other but not both. Chore will quit and make fu
 
 ### Tips for configuring Chore
 
+For Rails, it can be necessary to add the directory you store your jobs in to the eager loading path,
+found in application.rb. You likely need a similar approach for most apps using jobs, unless you places
+them into a directory that is already eager loaded by default. One example of this might be:
+
+```ruby
+config.eager_load_paths += File.join(config.root, "app", "jobs")
```
tannerburson: Should test whether this actually works. I don't think this works as expected in development mode.

StabbyCutyou: But thats a rails thing, isn't it? Eager Load is always different in dev
mode?

Happy to add a note around it, but this is taken almost verbatim (We have
more than this in our "add to EL path" code, so it's a bit more involved)
from our server code here at Tapjoy.

tannerburson: It is a rails thing, but it means if you follow this (and only this) Chore won't run correctly in dev mode. Historically we've had to actively load them in an initializer to ensure we have the classes loaded when we need them.

StabbyCutyou: Added a note about loading the jobs explicitly.

PR #15, opened 2015-12-09
First pass at doing a batched-worker mode (different from the batcher…

StabbyCutyou:
…) to coalesce multiple jobs into one for certain performance edgecases

ping @theo-lanman @Tapjoy/eng-group-services 

So, I took a stab at this. There is probably a little cleanup I code do, but I wanted to be as minimally invasive as possible, as Chore really isn't set up internally to make this kind of thing efficient - lots of stuff assumes one and only one message at a time (like the hooks, for example).

Theres lots we could do to improve that, but it's both outside the scope of this work, and not really the performance bottleneck we're aiming to solve (which is network latency in a very specific, high throughput job).

I looked at moving this behavior to the publisher / consumer to keep the worker agnostic, but the system just isn't designed for it overall, and we'd have a lot more to update. I can go over them in person better than explaining them here, but the gist is: Multiple processes writing to the same file where the data is likely above the linux limit on safe writes without interleaving data, the serialized job data would not be in a convenient format to match the downstream worker model, and it would have overall been more code to accomplish the same thing in what ended up looking like a more brittle approach.

It's a great use case to keep in mind if we get around to a proper design session on Chore 2.0.

Thoughts on this? Not yet tested, going to work on that shortly.

### DIFF
--- a/lib/chore/cli.rb
+++ b/lib/chore/cli.rb
@@ -141,6 +141,7 @@ module Chore #:nodoc:
 
       register_option 'dupe_on_cache_failure', '--dupe-on-cache-failure BOOLEAN', 'Determines the deduping behavior when a cache connection error occurs. When set to false, the message is assumed not to be a duplicate. (default: false)'
 
+      register_option 'process_in_batches', '--process-in-batches BOOLEAN', 'Determines if the Worker should hand each message to a single instance of a Job, or if the Jobs are intended to make multiple messages and work over all of them. (default: false)'
     end
 
     def parse_opts(argv) #:nodoc:
@@ -249,4 +250,3 @@ module Chore #:nodoc:
     end
   end
 end
-
--- a/lib/chore/manager.rb
+++ b/lib/chore/manager.rb
@@ -10,7 +10,10 @@ module Chore
       Chore.logger.info "Booting Chore #{Chore::VERSION}"
       Chore.logger.debug { Chore.config.inspect }
       @started_at = nil
-      @worker_strategy = Chore.config.worker_strategy.new(self)
+      worker_opts = {
+        :process_in_batches=>Chore.config.process_in_batches
+      }
+      @worker_strategy = Chore.config.worker_strategy.new(self, worker_opts)
       @fetcher = Chore.config.fetcher.new(self)
       @processed = 0
       @stopping = false
--- a/lib/chore/unit_of_work.rb
+++ b/lib/chore/unit_of_work.rb
@@ -8,7 +8,7 @@ module Chore
   # * +:previous_attempts+ The number of times the work has been attempted previously.
   # * +:consumer+ The consumer instance used to fetch this message. Most queue implementations won't need access to this, but some (RabbitMQ) will. So we
   # make sure to pass it along with each message. This instance will be used by the Worker for things like <tt>complete</tt> and </tt>reject</tt>.
-  class UnitOfWork < Struct.new(:id,:queue_name,:queue_timeout,:message,:previous_attempts,:consumer)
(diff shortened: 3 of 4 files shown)


gregorysabatino:
looks like any individual item in a batch that fails will trash the entire batch.  possible to expunge the offender (if there is one "bad" item that repeatedly fails) and allow the rest of the batch to proceed?

StabbyCutyou:
I don't think there is a great way to accomplish that with this approach - I think we'd need a larger re-architecting of how data is passed back from the running job to the chore process, as right now all we have is "Did it raise an error or not", which limits us to raising specific errors with info baked into them for the worker to try and interpret and take action over.

Limited as it is, I think the current behavior is good enough for the time being, until we see an issue.

StabbyCutyou:
@theo-lanman I made enough changes that if you had a PR open against this, I probably broke it.

But thats the risk I took when I pushed up untested code for a review :)

StabbyCutyou:
I'll squash all those little commits once we're happy with the whole approach overall

StabbyCutyou:
Squashed

obrie:
I think we'll want to be careful to consider some of the details with this approach.  With our Kafka implementation, we can only do batches per partition -- so if there are messages with mixed partitions in a single batch, we would need to consider what happens when half the batch is successful and half has failed.  Probably worth taking a step back and just making sure that batching aligns with the guarantees we want to have with publishing to Kafka.

StabbyCutyou:
We could account for that in the job itself, right?

Don't wanna mix application level concerns with chore concerns, but perhaps
a generic batching lambda could be defined in chore.rb, and applied to
batches.

But since we're shelving this for now, we can revisit the design later.

tannerburson:
I realize I'm late to the party... but what about an alternative approach here?

For starters, instead of adding more global configuration, why not make a subclass of Worker to contain this behavior? Combine that with a subclass of the appropriate worker strategy that uses your new BatchedWorker, and you've avoided a ton of configuration overhead, with no reduction in flexibility.

Secondly, if the performance profile of these jobs is so wildly different from the norm, why force it through the same master as everything else? It would seem that either a custom consumer that did per-queue batching, a sub-classed batcher doing per-queue batching, or possibly just a very short-timeout batcher would accomplish the same thing, more consistently, and with less additionally complexity.

StabbyCutyou:
We decided to shelve this for the time being, as we did some benchmarks and realized we didn't need it (yet).

The idea was to try to and see how little we could do to enable the functionality, just for quick testing. As I noted in the description, there are a lot of reasons why the current code isn't designed with this approach in mind, and the current implementation is a bit rough as it was just a proof of concept.

A subclassed worker isn't a good idea, I'll keep that in mind if we ever decide to actually put this approach into practice.

The reason we wanted to batch like this (hand an entire batch of messages into a single job), and not just do a per-queue batcher implementation, is for the specific type of job - it's basically just a middle-man between the local file system and a third party service which can accept batched-writes. Batching per-queue, without then handing the batch entirely to a single job, doesn't help us if the aim is to reduce the overall number of writes (and we cannot batch the data on the way into the queue) to improve the round-trip latency of the overall system.

I should likely close this for the time being as we're not currently moving forward with it, but I appreciate the design feedback.

tannerburson:
Ah, got it. Yeah, I'd cheat and do it in the Consumer, and maybe a
subclassed Worker to unwrap it. Or go full-cheat and do a Consumer and just
let the Job know about batches. But yeah, I saw this was a bit like raising
the dead, but wanted to throw my hat in the ring anyway.

StabbyCutyou:
Always appreciated.

Issue #5, opened 2015-10-16
Can't specify any class based args from the command line


tannerburson:
If you attempt to specify any of the class based args (consumer, publisher, strategies, etc), you can only use classes that exist inside of Chore. Because Chore doesn't attempt to call `Bundler.setup` or `.require` the only classes loaded are Chore itself. 

This makes it difficult to configure an app that is using a different queue implementation, without making the application itself aware of the change.

ping @StabbyCutyou

StabbyCutyou:
I'm guessing this is strictly the cmd line args, and not in something like a chore.rb?

StabbyCutyou:
In hindsight, the answer to my question was obvious.

I have a fix for this (I tested it with your disque consumer, it's able to load the constant at run time but then you have some downstream bugs that break things - however the CLI parsing is able to do it's thing).

I'd ping you on the actual PR to fix this but you recall the release process that we still use for this project. Once someone LGTMs it, i'll cut the gem for you and push it to rubygems.

tannerburson:
Thanks! I'll keep my eye out for the version bump.

tannerburson:
Fixed in the 1.7 release.

PR #17, opened 2015-12-15
Only explicitly use Bundler.require when in non-rails environments

obrie:
This is a regression introduced by https://github.com/Tapjoy/chore/issues/5.  Invoking `Bundler.require` within a Rails app *prior* to actually loading the rails libraries causes issues.  In particular:
* There are some gems that simply assume that Rails has been loaded (rather than require dependencies)
* More importantly, Railtie files typically only get loaded by gems when they detect that Rails is available (via something like `defined?(Rails)`).

By requiring all gems ahead of time, we break the above two assumptions since there's no guarantee on when Rails actually gets loaded.  We could explicitly define the `require` option for every gem, but that's going to be full of headaches.

Instead, this makes that behavior part of Chore's bootup process.  The primary reason for having this at all is so that 3rd-party chore gems can introduce additional cli options.  Since chore processes the cli options twice (once before boot and once after boot), this still works.

/to @StabbyCutyou 
/cc @Tapjoy/eng-group-services

### DIFF
--- a/bin/chore
+++ b/bin/chore
@@ -22,9 +22,6 @@ Chore::Signal.trap "USR1" do
 end
 
 begin
-  # Pre-load any Bundler dependencies now, so that the CLI parser has them loaded
-  # prior to intrpretting the command line args for things like consumers/producers
-  Bundler.require if defined?(Bundler)
   cli = Chore::CLI.instance
   cli.run!(ARGV)
 rescue => e
--- a/lib/chore/cli.rb
+++ b/lib/chore/cli.rb
@@ -63,10 +63,10 @@ module Chore #:nodoc:
       end
     end
 
-    def parse_config_file(file) #:nodoc:
+    def parse_config_file(file, ignore_errors = false) #:nodoc:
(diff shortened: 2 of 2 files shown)


StabbyCutyou:
I'm a little wary of the "Ignore Errors" behavior that seems to just lop off an argument from argv and try again... whats the reasoning there / intended use case?

obrie:
I'll clean that up so that it only retries arguments past the one that fails.

obrie:
@StabbyCutyou thoughts on the latest changes?

StabbyCutyou:
We spoke offline - with this change, we now support params that may not present on the first pass of the configuration run, but will be on the 2nd - so the "retry without error" is needed to help us smooth over the first run, then not used on 2nd.

With that in mind, that approach seems correct to me.

StabbyCutyou:
Seems good to me

obrie:
Updated with more comments

StabbyCutyou:
Marge it

PR #18, opened 2016-01-14
Add profiler

theo-lanman:
@Tapjoy/eng-group-services 

__DO NOT MERGE__

Here's the profiling code for Chore.

This includes a pre-release version tag. This will evaluate as an "earlier" version than 1.8.0, so this should be safe to build as a gem.

Does this look safe to build and run in production?

### DIFF
--- a/bin/chore
+++ b/bin/chore
@@ -1,6 +1,8 @@
 #!/usr/bin/env ruby
 
 $LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', 'lib'))
+require 'ruby-prof'
+RubyProf.start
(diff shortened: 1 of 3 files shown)


gregorysabatino:
LGTM

obrie:
Could we just add this to our app and have the Procfile reference our version of the binary?  I don't know if that'd be any easier than messing with the chore version.

Otherwise feel free to integrate this as a gem.

obrie:
(last statement should have included an lgtm)

theo-lanman:
Good idea about including it in the sidecar. Doing that.

PR #19, opened 2016-01-19
WIP: batched jobs with subclassed worker

theo-lanman:
@Tapjoy/eng-group-services 
@StabbyCutyou 

Here's the beginnings of an implementation of batched job handling using a subclassed worker. To use this, simply configure your chore consumer/worker process with the new batched strategy, and implement the `perform_batch` method on your jobs.

Looking for feedback on the approach here.

Todo:
- specs
- docs (comments and otherwise)
- partial failure handling (maybe `#perform_batch` should accept a map of job id => body, and return a report of job_id => status?)
- maybe some more refactoring around `#perform_batch`
- maybe `Job` should be subclassed to `BatchJob` with better primitives around partial failure/success handling?

### DIFF
--- a/lib/chore.rb
+++ b/lib/chore.rb
@@ -13,6 +13,7 @@ require 'chore/manager'
 require 'chore/publisher'
 require 'chore/util'
 require 'chore/worker'
+require 'chore/batched_worker'
 require 'chore/publisher'
 
 # We have a number of things that can live here. I don't want to track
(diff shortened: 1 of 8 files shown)


gregorysabatino:
makes sense to me

StabbyCutyou:
General feedback - is there a way to make the BatchedConsumer inherit more from the original worker, so we don't need to re-define the entire thing? Or do you think that makes it too complex to implement?

One thing I never liked about this was the overloading of the term "batch". We technically now have it twice - Once for the original use (the "batcher" in the forked worker) and now here. Worth the 5 minutes to think of a name that still expresses the same intent, but isn't overloading the term "batch"?

theo-lanman:
This is instead being written as a "plugin".

PR #20, opened 2016-01-21
Add support for configuring the consumer batch_timeout

theo-lanman:
@Tapjoy/eng-group-services 

This adds support for configuring the consumer `batch_timeout`. This is the maximum amount of time that the consumer will wait while consuming to complete a full batch of size `batch_size`. A batch is released to the worker when either `batch_size` is reached or `batch_timeout`. This is existing functionality; the only change is to build the last mile of track to the config.

### DIFF
--- a/lib/chore.rb
+++ b/lib/chore.rb
@@ -33,6 +33,7 @@ module Chore #:nodoc:
     :consumer              => Queues::SQS::Consumer,
     :fetcher               => Fetcher,
     :consumer_strategy     => Strategy::ThreadedConsumerStrategy,
+    :batch_timeout         => 20,
     :batch_size            => 50,
     :log_level             => Logger::WARN,
     :log_path              => STDOUT,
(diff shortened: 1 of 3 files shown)


gregorysabatino:
seeking to understand  (questions have nothing to do with the changes made, which LGTM): i'm not reading the code to behave as documented in the PR or in the code itself, or in the 'config option'?
in order for this setting to kick-in, `batch_timeout` must elapse since last message added to batch, not necessarily since last batch completed.  once timeout is reached, execute is immediately called (does not wait for next message to hit queue) ... although spirit of code documentation would be true if no messages at all hit queue until after `batch_timeout` expires.

theo-lanman:
Huh. On reading the code, I've come to the same conclusion.

I'm not sure whether we should A) change the behavior to match the documentation in the code, B) tweak the expected production config to deal with the existing behavior (and fix the comments/&c), or C) add a new feature that works the way I've described.

I'd love to do A, but that may have a negative effect on existing deployments. Let's talk tomorrow.

theo-lanman:
Spoke to @StabbyCutyou and @gregorysabatino in-person about this. The conclusion is that we're going to go with route B for now (find a happy-medium configuration that works in light of this behavior), and put A on the backlog.

PR #21, opened 2016-01-21
Memoize #methods in hook processing

theo-lanman:
hi @Tapjoy/eng-group-services 
cc @StabbyCutyou 

This improves performance significantly (when we're not waiting on network I/O).

This does involve a slight change of behavior: hook methods defined dynamically after jobs begin processing will no longer be called. I think this is reasonable.

### DIFF
--- a/lib/chore/hooks.rb
+++ b/lib/chore/hooks.rb
@@ -15,7 +15,13 @@ module Chore
 
   private
     def hooks_for(event)
-      (self.methods - Object.methods).grep(/^#{event}/).sort
+      candidate_methods.grep(/^#{event}/).sort
+    end
+
+    # NOTE: Any hook methods defined after this is first referenced (i.e.,
+    # after chore begins processing jobs) will not be called.
+    def candidate_methods
+      @_chore_hooks_candidate_methods ||= (self.methods - Object.methods)
     end
 
     def global_hooks_for(event)


StabbyCutyou:
I don't think dynamic hook creation after fork is an intended feature any
ways.

Will review tonight or tomorrow morning.

gregorysabatino:
if we believe use-case is possible, can we wipe out the memo when a job starts?

theo-lanman:
I'm with Stabby; I don't think dynamic hook addition is an intended feature. In any case, wiping out the memo when the job starts isn't useful, since this is only called a few times per job. The problem arises when we process hundreds-to-thousands of jobs per second.

gregorysabatino:
OK, not stuck on it.  Would just like to talk through your comment to understand it better.

StabbyCutyou:
Let's get a comment in here to explain: why this is being changed, what it impacted (dynamic hooks)

Any official numbers you can paste into the pr as proof of improvement would be a nice feather in the cap as well

That accomplished, lgtm

obrie:
This addresses the performance issues, so as-is lgtm, but also I'd be on board with just getting rid of this.  There's no need for this feature.  We should support calling a single method named exactly as the event is and call it a day.

However, I know that's additional work and would be a major version bump, so happy to ship this as well.

StabbyCutyou:
Thats also an idea that has merit

We should stick it in the "Chore 2.0" idea bucket.

theo-lanman:
Profiler results: in the unoptimized version, 47.0% of chore worker time was spent in `#run_hooks_for`. In the optimized (i.e., this) version, 9.6% of chore worker time was spent in `#run_hooks_for`. Doing the math, that means that batched sidecar jobs in the optimized version take 41.37% less time.

StabbyCutyou:
Yikes!

theo-lanman:
Comment added.

PR #22, opened 2016-01-21
Cache item klass and decoded_message

theo-lanman:
@Tapjoy/eng-group-services 

Sorry for the PR flood.

This is a little bit of refactoring necessary to support the batched worker as a chore "plugin".


gregorysabatino:
LGTM

PR #23, opened 2016-01-25
Log serializaiton errors

will-ob:
food_foodfactory_e2032 1

### DIFF
--- a/lib/chore/queues/filesystem/publisher.rb
+++ b/lib/chore/queues/filesystem/publisher.rb
@@ -24,6 +24,9 @@ module Chore
               if f.flock(File::LOCK_EX | File::LOCK_NB) && f.size == 0
                 begin
                   f.write(job.to_json)
+                rescue StandardError => e
+                  Chore.logger.error "#{e.class.name}: #{e.message}. Could not write #{job[:class]} job to #{queue_name} file.\n#{e.stacktrace}"
+                  raise e
                 ensure
                   f.flock(File::LOCK_UN)
                   break
--- a/lib/chore/queues/filesystem/publisher.rb
+++ b/lib/chore/queues/filesystem/publisher.rb
@@ -25,12 +25,11 @@ module Chore
                 begin
                   f.write(job.to_json)
                 rescue StandardError => e
-                  Chore.logger.error "#{e.class.name}: #{e.message}. Could not write #{job[:class]} job to #{queue_name} file.\n#{e.stacktrace}"
-                  raise e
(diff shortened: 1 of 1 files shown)


gregorysabatino:
that image is disturbing.
aside from that, LGTM, with one question : any use in logging the stack trace?

will-ob:
Ignore the cerealized-stream image.

I could go either way on the stack trace. On the one hand, it's nice to have since the error's origin is hidden in a third party library (chore). On the other hand, we re-raise the exception, so it needs to be handled elsewhere anyway.

lib/chore/queues/filesystem/publisher.rb
```diff
@@ -24,6 +24,9 @@ def publish(queue_name,job)
               if f.flock(File::LOCK_EX | File::LOCK_NB) && f.size == 0
                 begin
                   f.write(job.to_json)
+                rescue StandardError => e
+                  Chore.logger.warn "#{e.class.name}: #{e.message}. Could not write #{job[:class]} job to #{queue_name} file."
```
theo-lanman: Should this be logged as `error`?

will-ob: hmm... it iss an error, i suppose

lib/chore/queues/filesystem/publisher.rb
```diff
@@ -24,6 +24,9 @@ def publish(queue_name,job)
               if f.flock(File::LOCK_EX | File::LOCK_NB) && f.size == 0
                 begin
                   f.write(job.to_json)
+                rescue StandardError => e
+                  Chore.logger.error "#{e.class.name}: #{e.message}. Could not write #{job[:class]} job to #{queue_name} file.\n#{e.stacktrace}"
+                  raise e
```
theo-lanman: Also, I don't know if we want to raise here—that would be very new behavior, and I'm not sure if raising in the publishing code is always safe.

That said, I think we mostly use the filesystem queue in dev/tapinabox environments.

theo-lanman: It turns out that the exception is swallowed because we `break` in the `ensure` block below—the exception causes a jump to the ensure, but the break jumps out of the while and out of exception handling(!).

This _probably_ should have a `break` below the ensure block, but in order not to change behavior, we should remove the `raise` here.

Q_Q

theo-lanman:
I like this: now we're explicitly swallowing `StandardError`, and correctly _not_ swallowing other errors.

will-ob:
Alrighty, still good all? @gregorysabatino @theo-lanman

theo-lanman:
LGTM.

PR #24, opened 2016-01-28
If the single worker raises an exception, ensure that it is released

theo-lanman:
@Tapjoy/eng-group-services 
cc @StabbyCutyou 

We got another bad job file—and no more work was being processed, without error. This addresses that issue.

### DIFF
--- a/lib/chore/strategies/worker/single_worker_strategy.rb
+++ b/lib/chore/strategies/worker/single_worker_strategy.rb
@@ -21,13 +21,20 @@ module Chore
         worker.stop! if worker
       end
 
-      # Assigns work if there isn't already a worker in progress. Otherwise, is a noop
+      # Assigns work if there isn't already a worker in progress. In this, the
+      # single worker strategy, this should never be called if the worker is in
+      # progress.
       def assign(work)
         if workers_available?
-          @worker = Worker.new(work, @options)
-          @worker.start
-          @worker = nil
-          true
+          begin
+            @worker = Worker.new(work, @options)
+            @worker.start
+            true
+          ensure
+            @worker = nil
+          end
+        else
+          Chore.logger.error { "#{self.class}#assign: single worker is unavailable, but assign has been re-entered: #{caller * "\n"}" }
         end
       end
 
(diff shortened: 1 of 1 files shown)


StabbyCutyou:
You could drop line 29 if you wanna get picky, but otherwise I see the problem here. Lgtm.

obrie:
What happens in this scenario:
* Who catch and rescues the exception?
* Do filesystem jobs get recovered (i.e. made new again)?

StabbyCutyou:
The implementation in the forked worker is a little more thorough, but it's basically doing the same things here as far as I can see.

The worker bombing out because of exceptions should and re-trying the job should be handled elsewhere from what I remember.

obrie:
I had looked through the code and it wasn't obvious to me how this was being handled.  I'd like to make sure we look through the code and understand / test the behavior.  Whoever looks at it, please post references to lines of code that accomplish proper recovery.

StabbyCutyou:
Jobs that fail should be handled by the worker.rb itself (the base one in lib, not the strategies), but that brings up a good point - how did a crash in the worker itself bubble up to this point, when it should be handled by the worker itself to make sure the jobs get re-enqueued? Perhaps there are exceptions in worker.rb we're not accounting for?

theo-lanman:
An exception in `SingleWorkerStrategy#assign` is caught [here](https://github.com/Tapjoy/chore/blob/master/lib/chore/queues/filesystem/consumer.rb#L97). This causes the (consumer) batch being worked on to be abandoned in the `inprogress` directory, and the next batch will be attempted. Nothing is recovered.

It looks like the unaccounted-for exception that happened in this case is unique to the batched worker—the regular worker [catches this](https://github.com/Tapjoy/chore/blob/master/lib/chore/worker.rb#L59). It still abandons the job, but not the whole batch.

The right solution here is probably to log errors and recover as gracefully as possible in the non-batched worker—as it is, any error that bubbles up to here will cause all further processing to halt without any alert—but the solution for the batched worker is a little more involved.

theo-lanman:
Added some logging and cleaned this up a bit.

StabbyCutyou:
If something does re-enter this by accident, would the @worker=nil line break anything?

Just wanna make sure we've thought the ensure block all the way through, but in general I think this is probably better than nothing either way.

theo-lanman:
It sure would. How does this look?

StabbyCutyou:
Much safer :)

If this works and you're happy with it, LGTM

PR #25, opened 2016-02-01
Allow single worker strategy to be used in conjunction with multiple consumer threads

obrie:
This is a fix for the issue where having multiple consumer threads in conjunction with a single worker strategy can cause work to be left by the side because we don't have any locking around the worker.  This effectively imitates what's being done in the forked worker strategy.

I've decided to not do any refactoring around how some of the logic here is similar to the forked worker strategy.  I think there's a larger thought here around replacing the single worker strategy with a threaded worker strategy -- and getting us to a place where that shares much of the same code as the forked worker strategy.  That's a problem for another day, though.

In the meantime this should at least resolve this particular failure with multiple consumer threads.

/to @StabbyCutyou 
/fyi @Tapjoy/eng-group-services

### DIFF
--- a/lib/chore/strategies/worker/single_worker_strategy.rb
+++ b/lib/chore/strategies/worker/single_worker_strategy.rb
@@ -10,7 +10,10 @@ module Chore
       def initialize(manager, opts={})
         @options = opts
         @manager = manager
+        @stopped = false
         @worker = nil
+        @queue = Queue.new
+        @queue << :worker
       end
 
       # Starts the <tt>SingleWorkerStrategy</tt>. Currently a noop
@@ -18,6 +21,11 @@ module Chore
 
       # Stops the <tt>SingleWorkerStrategy</tt> if there is a worker to stop
       def stop!
+        return if @stopped
+
+        @stopped = true
+        Chore.logger.info { "Manager #{Process.pid} stopping" }
+
         worker.stop! if worker
       end
 
@@ -25,16 +33,14 @@ module Chore
       # single worker strategy, this should never be called if the worker is in
       # progress.
       def assign(work)
-        if workers_available?
-          begin
-            @worker = worker_klass.new(work, @options)
-            @worker.start
-            true
-          ensure
-            @worker = nil
-          end
-        else
-          Chore.logger.error { "#{self.class}#assign: single worker is unavailable, but assign has been re-entered: #{caller * "\n"}" }
+        return unless acquire_worker
+
+        begin
+          @worker = worker_klass.new(work, @options)
+          @worker.start
+          true
+        ensure
+          release_worker
         end
       end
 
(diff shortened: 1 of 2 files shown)


obrie:
I've verified through the chore tester that this fixes the described issue.

lib/chore/strategies/worker/single_worker_strategy.rb
```diff
@@ -25,16 +33,14 @@ def stop!
...
-          begin
-            @worker = worker_klass.new(work, @options)
-            @worker.start
-            true
-          ensure
-            @worker = nil
-          end
-        else
-          Chore.logger.error { "#{self.class}#assign: single worker is unavailable, but assign has been re-entered: #{caller * "\n"}" }
+        return unless acquire_worker
```
StabbyCutyou: Should this be more like the forked code, where we use an if block?

I get where acquire_worker will block, and thus this shouldn't ever actually "return" here, but it seems to _imply_ that you could return early if the worker is not available.

Just a thought

obrie: Not sure I follow -- as far as I can tell, this is what the forked code does (I effectively copied it from there).  Mind pointing to the lines of code (in the forked worker) you're thinking about?

StabbyCutyou: Might be a brain fart on my end, one sec...

StabbyCutyou: Ok, ignore me. For some reason I thought it was a "if acquire_worker"...
but it isn't!

LGTM

obrie: To clarify, are you LGTM'ing the whole PR or just this one line? :)

will-ob:
So we have one worker? As in, one process that actually performs work outside of this thread?

If that's the case, then this seems legit. Lgtm

lib/chore/strategies/worker/single_worker_strategy.rb
StabbyCutyou: Yes

StabbyCutyou: LGTM overall

theo-lanman:
LGTM.

StabbyCutyou:
LGTM

theo-lanman:
LGTM.

gregorysabatino:
don't leave me out!  : LGTM

StabbyCutyou:
LGTM

obrie:
LGTM

StabbyCutyou:
#deployable

will-ob:
This lgtm

theo-lanman:
I have some concerns about the approach here.

PR #26, opened 2016-02-02
Add before_start hook for running callbacks prior to starting consuming / processing

obrie:
This adds a `before_start` hook that mirrors the `before_shutdown` hook.  It will run callbacks prior to starting the fetcher / worker strategy (which are responsible for consuming and processing jobs).  This is particularly useful for integrations like NewRelic where we can have a single hook that works regardless of the worker strategy in use.  Currently, NewRelic integrations must hook into the `before_first_fork` hook which doesn't work if you're not using a forked strategy.

/to @StabbyCutyou @theo-lanman 
/cc @Tapjoy/eng-group-services

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,11 +189,13 @@ A number of hooks, both global and per-job, exist in Chore for your convenience.
 
 Global Hooks:
 
+* before_start
 * before_first_fork
 * before_fork
 * after_fork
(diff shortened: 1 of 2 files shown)


theo-lanman:
LGTM.

PR #27, opened 2016-02-03
Use multi_json instead of forcing usage of the json library

obrie:
`chore` currently requires that developers use the `json` library.  This makes it a bit more difficult to use alternative libraries like `oj`.  This updates chore's dependencies to use `multi_json` instead of `json`.  `json` remains a development dependency for testing purposes.

This will require a minor version bump.

/to @StabbyCutyou @theo-lanman

### DIFF
--- a/chore-core.gemspec
+++ b/chore-core.gemspec
@@ -36,11 +36,12 @@ Gem::Specification.new do |s|
   s.rubygems_version = "1.8.25"
   s.summary = "Job processing... for the future!"
 
-  s.add_runtime_dependency(%q<json>, [">= 0"])
+  s.add_runtime_dependency(%q<multi_json>, [">= 0"])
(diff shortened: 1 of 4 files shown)


StabbyCutyou:
By major, you mean this will officially be 2.0.0?

obrie:
Hah, I guess so?  I suspect if people were relying on this pulling in the `json` gem, then this could break their app...

obrie:
@StabbyCutyou verified, this should only be a minor version bump since JSON is baked into ruby 1.9+ core

StabbyCutyou:
So long as it's been tested, LGTM

obrie:
Closing this -- it's not the recommended approach anymore.  General recommendation in the community is to removal any dependency on json or multi_json and use the built-in json library in Ruby 1.9+.  If you want to use a different gem like `oj`, you need to use something like https://github.com/ohler55/oj_mimic_json.

PR #28, opened 2016-03-08
Improvements to filesystem queue handling

obrie:
Some proposals to make the way we publisher to / consumer from the filesystem queue to be consistent with other places in chore:
1. Encode jobs with `encode_job` rather than `to_json`
2. Don't swallow errors when writing to the job file
3. Encode the job before creating a job file so as to avoid the possibility of empty job files on encoding failures
4. Use `rm -f` when deleting job files

This should make it more consistent with the general way we handling jobs in sqs (we don't swallow errors, we use `encode_job`, etc.)

/to @StabbyCutyou @theo-lanman @gregorysabatino

### DIFF
--- a/lib/chore/queues/filesystem/consumer.rb
+++ b/lib/chore/queues/filesystem/consumer.rb
@@ -108,7 +108,7 @@ module Chore
 
         def complete(id)
           Chore.logger.debug "Completing (deleting): #{id}"
-          FileUtils.rm(File.join(@in_progress_dir, id))
+          FileUtils.rm_f(File.join(@in_progress_dir, id))
         end
 
         private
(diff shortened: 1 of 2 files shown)


gregorysabatino:
as we happy with where encoding / file generation errors will bubble up?

obrie:
FYI I've verified the behaviors locally.

theo-lanman:
Does this address errors when the file doesn't exist? That's one of the errors we see in production, right?

obrie:
I believe the `rm_f` will address it -- though I'm partly expecting that situation to not occur anymore with the changes to avoid blank files.

obrie:
@gregorysabatino I'm comfortable with it given how we integrate with it.  However, I think this'll need to be a major or minor version bump due to the fact that we could be raising exceptions in places we didn't previously (looking at @StabbyCutyou for guidance on the version bump here).

StabbyCutyou:
I'd be comfortable with a Minor bump - Major feels... too heavy handed?

gregorysabatino:
LGTM

PR #29, opened 2016-03-08
Reject failed jobs in the consumer

obrie:
We should be calling the `#reject` API in the consumer when a job fails.  As per the documentation, `#reject` "should put a message back on a queue to be processed again later".  When a job fails, this is exactly what we want to happen.  Now, in the SQS case this is a no-op because there's no concept for that type of functionality -- you just wait a little while and the job will be back in the queue.

However, in the filesystem case, you must explicitly put the job back into the queue.  Without this change, any job that raises an exception will be stuck in "inprogress" until chore is restarted.  With this change in place, the job will be moved back to "new" and ready to be re-processed.  That does mean that jobs will get *immediately* re-processes, but I think that's better than being stuck in inprogress forever.

More testing still needed.

/to @StabbyCutyou @theo-lanman @gregorysabatino

### DIFF
--- a/lib/chore/worker.rb
+++ b/lib/chore/worker.rb
@@ -66,6 +66,7 @@ module Chore
             item.consumer.complete(item.id)
           else
             Chore.run_hooks_for(:on_failure,item.message,e)
+            item.consumer.reject(item.id)
           end
         end
       end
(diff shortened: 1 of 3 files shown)


obrie:
This has been verified locally

StabbyCutyou:
SQS is already a noop for rejects, should have 0 impact on the existing chore w/r/t SQS, and help fix an issue on your end with the FSQ.

LGTM

PR #30, opened 2016-03-23
Change batch timeouts to be both configurable and based on the time oldest message

obrie:
This makes two changes:
1. The batch timeouts are now configurable instead of being hard-coded to 20s
2. We now run `execute` when the oldest message in the batcher has exceeded the timeout.

To provide a little context, the current behavior of the timeout is that it is based on the newest message in the batcher.  Consider the scenario where batch size = 10 and timeout = 1m.  If a message came in every 30s, then the batch wouldn't be processed until 5m has passed (because the batch size will then be 10).

I will also be looking for feedback / guidance on whether this should be a minor or major rev change.

/to @gregorysabatino @theo-lanman @StabbyCutyou

### DIFF
--- a/lib/chore/strategies/consumer/batcher.rb
+++ b/lib/chore/strategies/consumer/batcher.rb
@@ -11,27 +11,27 @@ module Chore
         @size = size
         @batch = []
         @mutex = Mutex.new
-        @last_message = nil
         @callback = nil
         @running = true
       end
 
       # The main entry point of the Batcher, <tt>schedule</tt> begins a thread with the provided +batch_timeout+ 
       # as the only argument. While the Batcher is running, it will attempt to check if either the batch is full, 
-      # or if the +batch_timeout+ has elapsed since the last batch was executed. If the batch is full, it will be executed.
-      # If the +batch_timeout+ has elapsed, as soon as the next message enters the batch, it will be executed.
+      # or if the +batch_timeout+ has elapsed since the oldest message was added. If either case is true, the
+      # items in the batch will be executed.
       # 
       # Calling <tt>stop</tt> will cause the thread to finish it's current check, and exit
-      def schedule(batch_timeout=20)
+      def schedule(batch_timeout)
         @thread = Thread.new(batch_timeout) do |timeout|
           Chore.logger.info "Batching timeout thread starting"
           while @running do
             begin 
-              Chore.logger.debug "Last message added to batch: #{@last_message}: #{@batch.size}"
-              if @last_message && Time.now > (@last_message + timeout)
-                Chore.logger.debug "Batching timeout reached (#{@last_message + timeout}), current size: #{@batch.size}"
+              oldest_item = @batch.first
+              timestamp = oldest_item && oldest_item.created_at
+              Chore.logger.debug "Oldest message in batch: #{timestamp}, size: #{@batch.size}"
+              if timestamp && Time.now > (timestamp + timeout)
+                Chore.logger.debug "Batching timeout reached (#{timestamp + timeout}), current size: #{@batch.size}"
                 self.execute(true)
-                @last_message = nil
               end
               sleep(1) 
             rescue => e
@@ -44,7 +44,6 @@ module Chore
       # Adds the +item+ to the current batch
       def add(item)
         @batch << item
-        @last_message = Time.now
         execute if ready?
       end
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Other options include:
     --threads-per-queue 4 # number of threads per queue for consuming from a given queue.
     --dedupe-servers # if using SQS or similiar queue with at-least once delivery and your memcache is running on something other than localhost
     --batch-size 50 # how many messages are batched together before handing them to a worker
+    --batch-timeout 20 # maximum number of seconds to wait until handing a message over to a worker
     --queue_prefix prefixy # A prefix to prepend to queue names, mainly for development and qa testing purposes
     --max-attempts 100 # The maximum number of times a job can be attempted
     --dupe-on-cache-failure # Determines the deduping behavior when a cache connection error occurs. When set to `false`, the message is assumed not to be a duplicate. Defaults to `false`.
@@ -92,6 +93,7 @@ Chore.configure do |c|
   c.max_attempts = 100
   ...
   c.batch_size = 50
+  c.batch_timeout = 20
 end
 ```
 
--- a/lib/chore.rb
+++ b/lib/chore.rb
@@ -34,6 +34,7 @@ module Chore #:nodoc:
     :fetcher               => Fetcher,
     :consumer_strategy     => Strategy::ThreadedConsumerStrategy,
     :batch_size            => 50,
+    :batch_timeout         => 20,
     :log_level             => Logger::WARN,
     :log_path              => STDOUT,
     :default_queue_timeout => (12 * 60 * 60), # 12 hours
@@ -186,6 +187,7 @@ module Chore #:nodoc:
   #   Chore.configure do |c|
   #     c.consumer = Chore::Queues::SQS::Consumer
   #     c.batch_size = 50
+  #     c.batch_timeout = 20
   #   end
   def self.configure(opts={})
     @config = (@config ? @config.merge_hash(opts) : Chore::Configuration.new(DEFAULT_OPTIONS.merge(opts)))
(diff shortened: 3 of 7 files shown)


theo-lanman:
Awesome! This looks like the right set of changes here.

My personal thought is that it should be a minor revision. Since batch timeout was not configurable before, but fixed at 20 seconds, at most this will cause systems which process batches _less often_ than every 20 seconds to process batches every 20 seconds.

theo-lanman:
Awesome! This looks like the right set of changes here.

My personal thought is that it should be a minor revision. Since batch timeout was not configurable before, but fixed at 20 seconds, at most this will cause systems which process batches _less often_ than every 20 seconds to process batches every 20 seconds.

lib/chore/strategies/consumer/batcher.rb
```diff
@@ -11,27 +11,27 @@ def initialize(size)
...
           while @running do
             begin 
-              Chore.logger.debug "Last message added to batch: #{@last_message}: #{@batch.size}"
-              if @last_message && Time.now > (@last_message + timeout)
-                Chore.logger.debug "Batching timeout reached (#{@last_message + timeout}), current size: #{@batch.size}"
+              oldest_item = @batch.first
+              timestamp = oldest_item && oldest_item.created_at
+              Chore.logger.debug "Oldest message in batch: #{timestamp}, size: #{@batch.size}"
+              if timestamp && Time.now > (timestamp + timeout)
+                Chore.logger.debug "Batching timeout reached (#{timestamp + timeout}), current size: #{@batch.size}"
```
StabbyCutyou: minor code comment, you could clean this up a bit like:

```ruby
if oldest_item = @batch.first && Time.now > (timestamp + timeout)
  Chore.logger...
  self.execute(true)
end
```

Not a blocker, just a nitpick

StabbyCutyou:
Only other comment is, should we be using Time.zone.now?

Otherwise, seems fine, non blocking nits aside.

lib/chore/strategies/consumer/batcher.rb
obrie: That doesn't work because `timestamp` is based on `oldest_item`.  Also, I've gotten the exact opposite feedback from other people in the past who don't like inline assignment in `if` statements so I tend to avoid it in more complex statements like this :)

StabbyCutyou: I updated the code to reflect where timestamp came from.

And that's just like, their opinion man.

Like I said, not a blocker.

obrie:
We shouldn't use `Time.zone.now` because that's an `active_support` thing.  Timezones are typically only interesting when you care about it from a visual perspective or where the underlying system has some dependency on either expecting or ignoring timezones.  In our case, we don't fall under any of those scenario.

StabbyCutyou:
Gotcha - I was thinking from the perspective of apps using chore, they'd have zone, but you're right - we lack it in chore itself. So, ignore that one.

lib/chore/strategies/consumer/batcher.rb
obrie: Understand that it's not a blocker, but I'm not just going to ignore the feedback.

In my opinion, inline assignment makes this less readable / clean because of the complexity of multiple conditions on the same line.  I rarely see that style used in that scenario.

The typical usage of inline assignment is:

```
if a = my_complex_thing
  do something with a
end
```

Not sure you want this PR to become a debate about inline assignment.  If that's the style you prefer for `chore`, I'll update this -- otherwise not sure your feedback is actionable.

StabbyCutyou: I don't want that :) I'm happy to have you leave it as is.

obrie:
Specs have been added and this has been validated locally.  See below:

Before:
```
[2016-03-23 15:32:06 -0400 (11465)] DEBUG : Found a new job development_report_metrics--20160323-193203-962908.0.job
[2016-03-23 15:32:06 -0400 (11465)] DEBUG : Got message: development_report_metrics--20160323-193203-962908.0.job
[2016-03-23 15:32:06 -0400 (11465)] DEBUG : Last message added to batch: 2016-03-23 15:32:06 -0400: 1
...
[2016-03-23 15:32:11 -0400 (11465)] DEBUG : Found a new job development_report_metrics--20160323-193210-787567.0.job
[2016-03-23 15:32:11 -0400 (11465)] DEBUG : Got message: development_report_metrics--20160323-193210-787567.0.job
[2016-03-23 15:32:11 -0400 (11465)] DEBUG : Last message added to batch: 2016-03-23 15:32:11 -0400: 2
...
[2016-03-23 15:32:16 -0400 (11465)] DEBUG : Found a new job development_report_metrics--20160323-193213-407804.0.job
[2016-03-23 15:32:16 -0400 (11465)] DEBUG : Got message: development_report_metrics--20160323-193213-407804.0.job
[2016-03-23 15:32:16 -0400 (11465)] DEBUG : Last message added to batch: 2016-03-23 15:32:16 -0400: 3
...
[2016-03-23 15:32:26 -0400 (11465)] DEBUG : Found a new job development_report_metrics--20160323-193223-412877.0.job
[2016-03-23 15:32:26 -0400 (11465)] DEBUG : Got message: development_report_metrics--20160323-193223-412877.0.job
[2016-03-23 15:32:26 -0400 (11465)] DEBUG : Last message added to batch: 2016-03-23 15:32:26 -0400: 4
...
[2016-03-23 15:32:31 -0400 (11465)] DEBUG : Found a new job development_report_metrics--20160323-193231-239298.0.job
[2016-03-23 15:32:31 -0400 (11465)] DEBUG : Got message: development_report_metrics--20160323-193231-239298.0.job
[2016-03-23 15:32:31 -0400 (11465)] DEBUG : Last message added to batch: 2016-03-23 15:32:31 -0400: 5
...
[2016-03-23 15:32:41 -0400 (11465)] DEBUG : Found a new job development_report_metrics--20160323-193238-869654.0.job
[2016-03-23 15:32:41 -0400 (11465)] DEBUG : Got message: development_report_metrics--20160323-193238-869654.0.job
[2016-03-23 15:32:41 -0400 (11465)] DEBUG : Last message added to batch: 2016-03-23 15:32:41 -0400: 6
...
[2016-03-23 15:32:51 -0400 (11465)] DEBUG : Found a new job development_report_metrics--20160323-193247-275925.0.job
[2016-03-23 15:32:51 -0400 (11465)] DEBUG : Got message: development_report_metrics--20160323-193247-275925.0.job
[2016-03-23 15:32:51 -0400 (11465)] DEBUG : Last message added to batch: 2016-03-23 15:32:51 -0400: 7
...
[2016-03-23 15:32:59 -0400 (11465)] INFO : Manager shutting down
```

After:
```
[2016-03-23 15:37:39 -0400 (17130)] DEBUG : Found a new job development_report_metrics--20160323-193737-432563.0.job
[2016-03-23 15:37:39 -0400 (17130)] DEBUG : Got message: development_report_metrics--20160323-193737-432563.0.job
[2016-03-23 15:37:40 -0400 (17130)] DEBUG : Oldest message in batch: 2016-03-23 15:37:39 -0400, size: 1
...
[2016-03-23 15:37:44 -0400 (17130)] DEBUG : Found a new job development_report_metrics--20160323-193742-258437.0.job
[2016-03-23 15:37:44 -0400 (17130)] DEBUG : Got message: development_report_metrics--20160323-193742-258437.0.job
[2016-03-23 15:37:45 -0400 (17130)] DEBUG : Oldest message in batch: 2016-03-23 15:37:39 -0400, size: 2
...
[2016-03-23 15:37:49 -0400 (17130)] DEBUG : Found a new job development_report_metrics--20160323-193749-280834.0.job
[2016-03-23 15:37:49 -0400 (17130)] DEBUG : Got message: development_report_metrics--20160323-193749-280834.0.job
[2016-03-23 15:37:50 -0400 (17130)] DEBUG : Oldest message in batch: 2016-03-23 15:37:39 -0400, size: 3
...
[2016-03-23 15:37:59 -0400 (17130)] DEBUG : Found a new job development_report_metrics--20160323-193755-007838.0.job
[2016-03-23 15:37:59 -0400 (17130)] DEBUG : Got message: development_report_metrics--20160323-193755-007838.0.job
[2016-03-23 15:38:00 -0400 (17130)] DEBUG : Oldest message in batch: 2016-03-23 15:37:39 -0400, size: 4
[2016-03-23 15:38:00 -0400 (17130)] DEBUG : Batching timeout reached (2016-03-23 15:37:59 -0400), current size: 4
[2016-03-23 15:38:00 -0400 (17130)] DEBUG : Manager#assign: No. of UnitsOfWork: 4)
```

obrie:
@StabbyCutyou Any thoughts on this being a minor rev bump instead of a major rev bump?

StabbyCutyou:
Mr Lanmans line of thinking is inline with my own - we made a fixed value configurable, and defaulted to the old fixed value.

Seems like a minor bump to me.

obrie:
@theo-lanman / @StabbyCutyou looking for a final review here

StabbyCutyou:
LGTM

theo-lanman:
This looks good to me. Yay specs!

PR #31, opened 2016-04-15
Preforked worker and throttled consumer strategies

goose42:
Objective
==
The main objective of the changes in this PR is to reduce job latencies on chore. We define latency on chore, as the time duration from when the message is consumed from a source, to when it is completed by a worker. The metric that we use to measure the latency is the average job latency per second. 

We aim to meet this objective with the introduction of these two strategies: 

- Pre-Forked worker strategy
- Throttled consumer strategy

Pre-Forked worker strategy: 
--
The goals of the pre-forked worker strategy are:
 - Manage the lifecycle of a group of workers. 
 - Manage job assignments to workers to minimize latency

Throttled consumer:
--
The goals of the throttled consumer strategy are:
 - Minimize the jobs in the chore which are in-flight but not being worked on. 
 - Always keep the internal pipeline of jobs full.

Tests and observations:
--
To test the effectiveness of the changes in the PR, we used this test job:
```
def perform(*args)
    sleep(1)
    5000.times{ |i| Digest::MD5.hexdigest(JSON.generate({ i => args})) }
end
```

We set up 2 servers to execute chore, one running the pre-forked worker strategy with the throttled consumers strategy and the other one running the forked worker strategy and with the threaded consumer.  The chore instances were running on Ruby 2.3.0 (MRI)

The configurations for both cases were: 
```
c.publisher           = Chore::Queues::SQS::Publisher
c.consumer            = Chore::Queues::SQS::Consumer

# note: batch_size only relevant for the threaded consumer strategy
c.batch_size          = 10
c.threads_per_queue   = 2
c.dedupe_servers      = 'localhost:11211'

c.num_workers         = 10
c.max_attempts        = 1000
c.default_queue_timeout = 60

c.queue_polling_size  = 10
```

Over a 1 hour run, this we measured the average latency of each job that was processed. We observed that the average latency was consistently, and considerably lower in the new case of the the new strategies. 
Comparing latency

In addition to measuring latency, we also measured the throughput (i.e. total number of jobs completed per second). While our primary objective was to reduce latency, we wanted to ensure that we would not affect the throughput in any drastic way. Our observation shows that the throughput over a 1 hour window is more or less similar between the two sets of strategies. 
Comparing Throughput

The memory consumption was also measured across the two servers over a one hour period. This is the memory used on the server running the preforked worker strategy:
Preforked worker memory used

This is the memory used on the server running the forked worker strategy:
Forked worker memory used

Additional constraints:
--
The only constraint imposed is that, the pre-forked worker strategy and throttled consumer strategy can only be used with each other; They are not intended to be used with any other worker or consumer strategies.

Please review: @StabbyCutyou @surendrapathak @obrie @Ceraunograph @theo-lanman @andyleclair @marcuswalser @riteshnoronha @alvindu @pravinchandruTapjoy

### DIFF
--- a/lib/chore/cli.rb
+++ b/lib/chore/cli.rb
@@ -89,6 +89,7 @@ module Chore #:nodoc:
       detect_queues
       Chore.configure(options)
       Chore.configuring = false
+      validate_strategy!
     end
 
 
@@ -254,7 +255,6 @@ module Chore #:nodoc:
     end
 
     def validate! #:nodoc:
-
       missing_option!("--require [PATH|DIR]") unless options[:require]
 
       if !File.exist?(options[:require]) ||
@@ -266,7 +266,25 @@ module Chore #:nodoc:
         puts @parser
         exit(1)
       end
+    end
 
+    def validate_strategy!
+      consumer_strategy = Chore.config.consumer_strategy.to_s
+      worker_strategy = Chore.config.worker_strategy.to_s
+
+      throttled_consumer = 'Chore::Strategy::ThrottledConsumerStrategy'
+      preforked_worker = 'Chore::Strategy::PreForkedWorkerStrategy'
+
+      if consumer_strategy == throttled_consumer || worker_strategy == preforked_worker
+        unless consumer_strategy == throttled_consumer && worker_strategy == preforked_worker
+          puts "=================================================================="
+          puts "  PreForkedWorkerStrategy may only be paired with   "
+          puts "  ThrottledConsumerStrategy or vice versa  "
+          puts "  Please check your configurations "
+          puts "=================================================================="
+          exit(1)
+        end
+      end
     end
   end
 end
--- a/lib/chore/manager.rb
+++ b/lib/chore/manager.rb
@@ -5,14 +5,15 @@ require 'chore/fetcher'
 module Chore
   # Manages the interactions between fetching messages (Consumer Strategy), and working over them (Worker Strategy)
   class Manager
+    include Util
 
     def initialize()
       Chore.logger.info "Booting Chore #{Chore::VERSION}"
       Chore.logger.debug { Chore.config.inspect }
+      procline("chore-master-#{Chore::VERSION}:Started:#{Time.now}")
       @started_at = nil
       @worker_strategy = Chore.config.worker_strategy.new(self)
       @fetcher = Chore.config.fetcher.new(self)
-      @processed = 0
       @stopping = false
     end
 
@@ -42,7 +43,15 @@ module Chore
     # than they can be consumed.
     def assign(work)
       Chore.logger.debug { "Manager#assign: No. of UnitsOfWork: #{work.length})" }
+      work.each do | item |
+        Chore.run_hooks_for(:before_send_to_worker, item)
+      end
       @worker_strategy.assign(work) unless @stopping
+    end
+
+    # returns up to n from the throttled consumer queue
+    def fetch_work(n)
+      @fetcher.provide_work(n)
     end
   end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -13,3 +13,4 @@ group :development do
 end
 
 gem 'ci_reporter', '~> 1.9.0'
+gem 'get_process_mem', '~> 0.2.0'
--- a/chore-core.gemspec
+++ b/chore-core.gemspec
@@ -39,6 +39,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency(%q<json>, [">= 0"])
   s.add_runtime_dependency(%q<aws-sdk-v1>, ["~> 1.56", ">= 1.56.0"])
   s.add_runtime_dependency(%q<thread>, ["~> 0.1.3"])
+  s.add_runtime_dependency('get_process_mem', ["~> 0.2.0"])
   s.add_development_dependency(%q<rspec>, ["~> 3.3.0"])
   s.add_development_dependency(%q<rdoc>, ["~> 3.12"])
   s.add_development_dependency(%q<bundler>, [">= 0"])
--- a/lib/chore.rb
+++ b/lib/chore.rb
@@ -41,7 +41,7 @@ module Chore #:nodoc:
     :shutdown_timeout      => (2 * 60),
     :max_attempts          => 1.0 / 0.0, # Infinity
     :dupe_on_cache_failure => false,
-    :payload_handler => Chore::Job
+    :payload_handler       => Chore::Job
   }
 
   class << self
@@ -111,9 +111,9 @@ module Chore #:nodoc:
   #   add_hook(:before_fork) {|worker| puts 1 }
   #   add_hook(:before_fork) {|worker| puts 2 }
   #   add_hook(:before_fork) {|worker| puts 3 }
-  #   
+  #
   #   run_hooks_for(:before_fork, worker)
-  #   
+  #
   #   # ...will produce the following output
   #   => 1
   #   => 2
@@ -130,9 +130,9 @@ module Chore #:nodoc:
   #   add_hook(:around_fork) {|worker, &block| puts 'before 1'; block.call; puts 'after 1'}
   #   add_hook(:around_fork) {|worker, &block| puts 'before 2'; block.call; puts 'after 2'}
   #   add_hook(:around_fork) {|worker, &block| puts 'before 3'; block.call; puts 'after 3'}
-  #   
+  #
   #   run_hooks_for(:around_fork, worker) { puts 'block' }
-  #   
+  #
   #   # ...will produce the following output
   #   => before 1
   #   => before 2
--- a/lib/chore/consumer.rb
+++ b/lib/chore/consumer.rb
@@ -53,5 +53,10 @@ module Chore
     def running?
       @running
     end
+
+    # returns up to n work
+    def provide_work(n)
+      raise NotImplementedError
+    end
   end
 end
--- a/lib/chore/fetcher.rb
+++ b/lib/chore/fetcher.rb
@@ -33,5 +33,10 @@ module Chore
     def stopping?
       @stopping
     end
+
+    # returns upto n work units
+    def provide_work(n)
+      @strategy.provide_work(n)
+    end
   end
 end
--- a/lib/chore/queues/sqs/consumer.rb
+++ b/lib/chore/queues/sqs/consumer.rb
@@ -116,9 +116,7 @@ module Chore
         def sqs
           @sqs ||= AWS::SQS.new(
             :access_key_id => Chore.config.aws_access_key,
-            :secret_access_key => Chore.config.aws_secret_key,
-            :logger => Chore.logger,
-            :log_level => :debug)
+            :secret_access_key => Chore.config.aws_secret_key)
         end
 
         def sqs_polling_amount
--- a/lib/chore/strategies/consumer/threaded_consumer_strategy.rb
+++ b/lib/chore/strategies/consumer/threaded_consumer_strategy.rb
@@ -64,6 +64,7 @@ module Chore
               Chore.logger.debug { "Got message: #{id}"}
 
               work = UnitOfWork.new(id, queue_name, queue_timeout, body, previous_attempts, consumer)
+              Chore.run_hooks_for(:consumed_from_source, work)
               @batcher.add(work)
             end
           rescue Chore::TerribleMistake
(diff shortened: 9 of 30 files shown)


StabbyCutyou:
The graphs are awesome to see

But one thing i'd point out, the memory consumption (while important to see in a "vanilla" sense), likely don't tell the full story of a job running in our primary service, which we know has a good deal of memory bloat.

It'd be good to see a sample of the memory graphs running a few select jobs. I can help you locate the right jobs outside of the PR, since they're company-specific.

Really cool to see this coming to fruition, looking forward to reviewing the code.

lib/chore/manager.rb
```diff
@@ -5,14 +5,15 @@
 module Chore
   # Manages the interactions between fetching messages (Consumer Strategy), and working over them (Worker Strategy)
   class Manager
+    include Util
 
     def initialize()
       Chore.logger.info "Booting Chore #{Chore::VERSION}"
       Chore.logger.debug { Chore.config.inspect }
+      procline("chore-master-#{Chore::VERSION}:Started:#{Time.now}")
```
andyleclair: A+

obrie:
A couple of things that would be super helpful in understand the high-level design before digging into the details:

1. What kind of review are you looking for here?  Are you looking for high-level design feedback, low-level bugs, external dependencies, etc?  Your answer here might impact the need for thinking through some of the next few items.
2. The PR gives a great description of the current state, but it doesn't really tell the story of the decisions we made to get to this point.  It'd be really helpful to understand more of the design of the elements here.  For example, Pre-forked vs. Threaded, Sockets vs. Pipes, Throttled vs. Non-Throttled
3. What do the interactions look like between the manager, consumer, worker, distributor, worker *manager*, sockets, etc.  There are lots of new concepts here and I think it could make reviewing this easier if we had, for example, a diagram showing the ways in which they're associated and what the interactions are between them (like a workflow-type diagram).
4. Any planned follow-on work that you intend to do, but isn't part of this first milestone (e.g. perhaps you're planning on making pre-forked the standard or maybe making the throttled strategy work for non-pre-forked strategies, etc.)
5. It'd be great to understand a little bit more of the testing strategy (e.g. maybe you have a test matrix or doc that outlines all of the scenarios that needed to be tested)
6. If we don't intent on using pre-forked everywhere, guidance on when we would use it.

There are also other things that it'd be great to understand a little more of the context for.  For example:
1. What are the new hooks we've added and how are they intended to be used?
2. Reasons for subtle changes like removing some of the options that we pass to SQS

Lots of really good stuff to review here, but I want to make sure I've got the full context of the design in my head and what you're looking to get out of the review before digging too deep.  Maybe there's enough stuff here that it's worth reviewing over VC (just a thought, not suggesting it's needed).

StabbyCutyou:
On thing I was thinking about on my way in this morning - this is probably a case where an initial, offline review might be helpful. There are so many new pieces and new behaviors introduced here, that while we could all walk through them piece by piece, having you guys guide us through the high level changes and new concepts would go a long way toward us giving this an accurate review.

We should then take the feedback and notes from the meeting and coalesce them here in the PR, so anyone else monitoring the thread can benefit from the effort.

lib/chore/manager.rb
StabbyCutyou: This will mean a change to some of our monitoring, as we look for a specific procline to determine the master.

That said, highly approve of this change.

andyleclair:
paging @tannerburson

goose42:
@obrie 

>What kind of review are you looking for here? Are you looking for high-level design feedback, low-level bugs, external dependencies, etc? Your answer here might impact the need for thinking through some of the next few items.

Mostly high level design reviews. We are working on workflow diagrams which would perhaps help better explain this implementation. We will present it soon.

>The PR gives a great description of the current state, but it doesn't really tell the story of the decisions we made to get to this point. It'd be really helpful to understand more of the design of the elements here. For example, Pre-forked vs. Threaded, Sockets vs. Pipes, Throttled vs. Non-Throttled

The main motivation of all these changes are to improve rewarding latencies. We found that the batch concept does not really work for latency sensitive jobs especially since setting a batch size of 1, 2, or 3 is not feasible (when forking). We had just the following design goals when we started out:
- Only have the minimum number of users jobs in the chore instance,which can be worked on. 
- Jobs which cannot be worked on immediately should be in SQS, add more instances to process these jobs 
So any design decision that was made, was made with those goals in mind. 

That being said, a detailed design document could be useful. I can start working on that. 

>What do the interactions look like between the manager, consumer, worker, distributor, worker manager, sockets, etc. There are lots of new concepts here and I think it could make reviewing this easier if we had, for example, a diagram showing the ways in which they're associated and what the interactions are between them (like a workflow-type diagram).

I am working on this bit of documentation right now. I will share it as soon as it is ready.

>Any planned follow-on work that you intend to do, but isn't part of this first milestone (e.g. perhaps you're planning on making pre-forked the standard or maybe making the throttled strategy work for non-pre-forked strategies, etc.)

The only follow on work have planned are:
- Build monitoring tools for operations, to better help them tune pre-fork chore
- "Smart configurations" for preforked workers, which could allow for automatic configuration of consumers & workers, and automatic scaling ensure that an instance uses it's resources optimally 

>It'd be great to understand a little bit more of the testing strategy (e.g. maybe you have a test matrix or doc that outlines all of the scenarios that needed to be tested)

I am currently working on an integration testing document, which I will be able to share soon. 

>If we don't intent on using pre-forked everywhere, guidance on when we would use it.

The main intention of the pre-forked worker was to reduce reward latencies. So, all the design decisions have been towards that objective. From that perspective, if you have a job that is latency sensitive, you should use this strategy. 
>What are the new hooks we've added and how are they intended to be used?

The new hooks are primarily intended for measuring the inner workings of this strategy. i.e. If a message is delayed, we would like to see exactly where in the flow of it, it is waiting. 
Here is an example of one such measurement:
chore_measurements

>Reasons for subtle changes like removing some of the options that we pass to SQS

We have only disabled debug logging. Not sure why this is enabled in production, but is there a reason it should be?

In terms of having a meeting to go over design details, I think that would be great idea. We should perhaps do that as soon as possible. :smiley:

StabbyCutyou:
On the subject of "smart" workers that autoconfig, you might want to talk
to geoff. Hes working on a way to do autoconfig at a high level based on
the instance type of the box it's run on.

As for debug logging... no one ever regretted the option for more details
when needed ;)
On Apr 16, 2016 12:47 AM, "Visweswaran Jayanth" <notifications@github.com>
wrote:

@obrie <https://github.com/obrie>

What kind of review are you looking for here? Are you looking for
high-level design feedback, low-level bugs, external dependencies, etc?
Your answer here might impact the need for thinking through some of the
next few items.

Mostly high level design reviews. We are working on workflow diagrams which
would perhaps help better explain this implementation. We will present it
soon.

The PR gives a great description of the current state, but it doesn't
really tell the story of the decisions we made to get to this point. It'd
be really helpful to understand more of the design of the elements here.
For example, Pre-forked vs. Threaded, Sockets vs. Pipes, Throttled vs.
Non-Throttled

The main motivation of all these changes are to improve rewarding
latencies. We found that the batch concept does not really work for latency
sensitive jobs especially since setting a batch size of 1, 2, or 3 is not
feasible (when forking). We had just the following design goals when we
started out:

   - Only have the minimum number of users jobs in the chore instance,which
   can be worked on.
   - Jobs which cannot be worked on immediately should be in SQS, add more
   instances to process these jobs So any design decision that was made, was
   made with those goals in mind.

That being said, a detailed design document could be useful. I can start
working on that.

What do the interactions look like between the manager, consumer, worker,
distributor, worker manager, sockets, etc. There are lots of new concepts
here and I think it could make reviewing this easier if we had, for
example, a diagram showing the ways in which they're associated and what
the interactions are between them (like a workflow-type diagram).

I am working on this bit of documentation right now. I will share it as
soon as it is ready.

Any planned follow-on work that you intend to do, but isn't part of this
first milestone (e.g. perhaps you're planning on making pre-forked the
standard or maybe making the throttled strategy work for non-pre-forked
strategies, etc.)

The only follow on work have planned are:

   - Build monitoring tools for operations, to better help them tune
   pre-fork chore
   - "Smart configurations" for preforked workers, which could allow for
   automatic configuration of consumers & workers, and automatic scaling
   ensure that an instance uses it's resources optimally

It'd be great to understand a little bit more of the testing strategy (e.g.
maybe you have a test matrix or doc that outlines all of the scenarios that
needed to be tested)

I am currently working on an integration testing document, which I will be
able to share soon.

If we don't intent on using pre-forked everywhere, guidance on when we
would use it.

The main intention of the pre-forked worker was to reduce reward latencies.
So, all the design decisions have been towards that objective. From that
perspective, if you have a job that is latency sensitive, you should use
this strategy.

What are the new hooks we've added and how are they intended to be used?

The new hooks are primarily intended for measuring the inner workings of
this strategy. i.e. If a message is delayed, we would like to see exactly
where in the flow of it, it is waiting.
Here is an example of one such measurement:
[image: chore_measurements]
<https://cloud.githubusercontent.com/assets/615316/14579125/dff60f32-0352-11e6-8bcf-a91ac8874bf0.png>

Reasons for subtle changes like removing some of the options that we pass
to SQS

We have only disabled debug logging. Not sure why this is enabled in
production, but is there a reason it should be?

In terms of having a meeting to go over design details, I think that would
be great idea. We should perhaps do that as soon as possible. [image:
:smiley:]

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/Tapjoy/chore/pull/31#issuecomment-210734679>

lib/chore/cli.rb
```diff
@@ -266,7 +266,25 @@ def validate! #:nodoc:
         puts @parser
         exit(1)
       end
+    end
 
+    def validate_strategy!
```
tannerburson: @StabbyCutyou This is another point towards needing to refactor the config support. Clunky that adding a set of strategies requires modifying the CLI class to handle validation cleanly.

StabbyCutyou: True - validation of the config is never something we really got quite
right.

tannerburson:
Obviously I'm lacking in the context, or details of the discussions that have been had internally, but I've got a few high level thoughts. I haven't had a chance to pull this down and dig through it locally, so these comments are just based on poking through it visually on github.

There's a ton of work here, and it solves for a number of problems we had early on when attempting to do something similar with pipe based communication. The overhead of managing the pipes for each child eventually swamped the manager, but unix sockets probably alleviate most of that, which is super cool to see.

This setup as a whole seems to break the loose guarantees that Chore attempts to setup around the flow of data from Queue -> Worker. That's evident in the need to add a pull mechanism (`fetch_work`) into each layer. The advantage to the existing workflow was always that it was "self-throttling" by making the effectively blocking call to `assign` from the fetching tier. It should be possible to make a batcher-less threaded consumer that still kept the "push" semantics of assign and combine it with the pre-forked worker strategy. It's similar to how the earliest versions of Chore were designed, and would simplify the fetching end of this dramatically.

That said, we always recognized that if absolute latency were the key, Chore's model wouldn't be the best option. It's always been a general purpose solution, and therefore unable to meet every need. If latency through the queue were the most important metric, I'd actually skip most of this and go to a much, much simpler model (which you could probably hack together using Chore plugins anyway). Effectively, you want to combine the fetcher and worker into a single unit so that each message is processed as fetched, with no waiting (or by building a small internal backlog to process, similar to how the current threaded consumer handles SQS batches). You could probably simulate this pretty easily with a modified single consumer strategy, and a bunch of master processes (effectively single process master/worker combo). It would have some overhead in the overall funnel through Chore, but wouldn't require the IPC gymnastics, nor complex child process management as you're now pushing that up to Upstart or whatever.

I'm not sure if that addresses the exact same concerns that you have, but it might be a simpler to integrate/operate option that requires less fighting against the grain of Chore's model, but still puts your fetcher/worker in closer proximity.

StabbyCutyou:
The quick overview of context is that we are, in fact, trying to tune both for absolute latency, as well as performance. Forking is still costly.

I think the issue with the fetcher + worker model is that there is no model for using Chore with our primary service at the moment other than to parallelize via forking. Therefore, for the fetcher+worker model to work, it could only be threadsafe jobs (in the current model), which we know we're still not confident about. (Unless i'm misunderstanding you).

I have been angling to get us to focus on a threaded worker model (Like the one you left in PR prior to leaving, which I still think is a good idea), but we'd be limited in the number of jobs we could apply it to out of the box.

The "unicorn" style worker, conceptually at least, works for all jobs out of the gate, but greatly reduces the performance cost of the fork.

Thats the high level rundown, anyways.

tannerburson:
Makes sense. Not a surprising place to end up at all. You're right, that obviously threaded workers would be ideal, but still complex in a lot of cases.

My suggestion was to cheat (as most of my suggestions are), and not spin up children at all. Run multi-master where each master is a single process fetcher/worker combo (probably just modified single worker strategy). Since you're probably only latency bound on a small number of queues, this shouldn't be too difficult to configure.

StabbyCutyou:
That's probably also something worth considering.

PR #33, opened 2016-05-20
added worker request/memory based killer

Ceraunograph:
@goose42 @riteshnoronha 

Added a memory/request based worker killer to avoid memory leaks.

Please Review

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,3 +13,4 @@ group :development do
 end
 
 gem 'ci_reporter', '~> 1.9.0'
+gem 'get_process_mem', '~> 0.2.0'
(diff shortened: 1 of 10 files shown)


Gemfile
```diff
@@ -13,3 +13,4 @@ group :development do
 end
 
 gem 'ci_reporter', '~> 1.9.0'
+gem 'get_process_mem'
```
goose42: Tie this down to a specific version?

lib/chore/strategies/worker/helpers/preforked_worker.rb
```diff
@@ -16,6 +16,8 @@ def initialize(_opts = {})
         @consumer_cache = {}
         @running = true
         post_fork_setup
+
+        @worker_killer = WorkerKiller.new
```
goose42: Doesn't need to be an instance variable.

riteshnoronha:
lgtm

PR #35, opened 2016-06-09
After fork hook

goose42:
Adding an after fork hook to the pre-forked worker.

@riteshnoronha @obrie @StabbyCutyou @Ceraunograph

### DIFF
--- a/lib/chore/strategies/worker/helpers/preforked_worker.rb
+++ b/lib/chore/strategies/worker/helpers/preforked_worker.rb
@@ -104,6 +104,7 @@ module Chore
         # It is possible for this to be nil due to configuration woes with chore
         config.publisher.reset_connection! if Chore.config.publisher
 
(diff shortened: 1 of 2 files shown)


StabbyCutyou:
Makes sense

riteshnoronha:
@goose42 can we audit any other hooks that we missed ?

goose42:
@riteshnoronha yeah, working on that list.

goose42:
Looks like we missed these hooks for the pre-forked worker: `around_fork`, `within_fork`, `after_fork`, `before_fork_shutdown`. The absent hooks will be added with a later PR (currently we only use the `after_fork`).

riteshnoronha:
@goose42 this works. I verified this in production. However we should review all other hooks, since it appears that worker objects are being passed around and used in plugins.. which is not really ideal.. 

lgtm

PR #36, opened 2016-06-16
USR1 signal handler

goose42:
Adding handlers for USR1 signal to the preforked worker strategy.

Also, changing some logs levels from debug to info. 

For review: @riteshnoronha @alanbrent @Ceraunograph @obrie @StabbyCutyou


riteshnoronha:
@goose42 and myself verified that USR1 is being handled correctly. 

lgtm

PR #37, opened 2016-06-29
add exception logging to help debugging

kchard:
@Tapjoy/tools @riteshnoronha @goose42 @Ceraunograph 

When trying to debug why some clicks where not being processed in Otto, I noticed that the WorkDistributor was swallowing exceptions and not providing any logging. This made debugging very difficult. Let's add some exception info.


goose42:
lgtm

PR #38, opened 2016-07-12
Enable travis CI testing

goose42:
Adding a .travis.yml file to enable tests to run via Travis CI.

For review: @maudineormsby @andyleclair

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: 
+  - ruby
+rvm: 
+  - 2.3.0
+script: 
+  - bundle exec rspec
+git:
+  depth: 10
+cache:
+  - bundle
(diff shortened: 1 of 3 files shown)


andyleclair:
@goose42 can you push another dummy commit? I think I have travis configured correctly, but it won't trigger until there's another commit pushed

andyleclair:
Now it's running!

andyleclair:
@goose42 There's a spec failure there, can you look into that?

andyleclair:
You should add a couple of things here to speed up the spec runs:

```
git:
  depth: 10
```
Only check out the last 10 commits, so we don't have to clone the whole git history

```
cache:
  - bundle
```
So Travis can keep the bundle around, and only reinstall if the bundle changes

goose42:
@andyleclair Specs should be fixed now. Also updated the travis config to speed up spec runs.

andyleclair:
ooh baby that's the good stuff.

LGTM

PR #40, opened 2017-01-06
[FIX] Close sockets upon worker death

Ceraunograph:
@goose42 @riteshnoronha 

Small fix to close sockets when the worker is killed and reaped.


goose42:
@Ceraunograph Can we also add a log line [here](https://github.com/Tapjoy/chore/blob/master/lib/chore/strategies/worker/helpers/preforked_worker.rb#L13), that says something on the lines of `"Worker #{Process.pid} initializing"`?

Ceraunograph:
Done

goose42:
LGTM. 👍

PR #41, opened 2017-03-22
Better logging during shutdown and prevent shutdown race condition.

riteshnoronha:
@Ceraunograph @goose42 

This fixes the issue of the master process hanging around on shutdown. 

cc: @marcuswalser @alanbrent


betamatt:
I'm not in a position to comment on the wisdom of forcing the process kill while it has threads trying to quiesce but an ensure couldn't hurt, unless you can guarantee that `worker_assignment_loop` can't throw anything else.

```
def worker_assignment_thread
  Thread.new do
    begin
      worker_assignment_loop
    rescue Chore::TerribleMistake => e
      Chore.logger.error 'PWS: Terrible mistake, shutting down Chore'
      Chore.logger.error e.message
      Chore.logger.error e.backtrace
    ensure
      Chore.logger.info "PWS: worker_assignment_thread ending"
      # WorkerAssignment thread is independent of the main thread.
      # The main thread is waiting on the consumer threads to join,
      # Due to some weird SQS behaviour, its possible that these threads
      # maynot join, and the assigment thread always exits, since it's 
      # nonblocking. This will ensure that the master process exits.
      Process.exit!(true)
    end    
  end
end
```

marcuswalser:
Sup @betamatt!

lib/chore/strategies/worker/preforked_worker_strategy.rb
```diff
@@ -23,6 +23,7 @@ def initialize(manager, opts = {})
         @self_read, @self_write = IO.pipe
         trap_signals(NUM_TO_SIGNAL, @self_write)
         @worker_manager = WorkerManager.new(create_master_socket)
+        at_exit { delete_socket_file}
```
riteshnoronha: @Ceraunograph should prevent orphaned socket files

Ceraunograph: It seem fine in theory, although I think it obscures shutdown behavior. Has this been tested to work?

goose42:
LGTM. 👍

PR #42, opened 2017-08-15
Memoize hook lookups to reduce performance impact when processing large numbers of jobs

obrie:
This memoizes the candidate lookup for hooks so that we only do the search once per class.  Without this, there's a pretty significant amount of overhead when processing very small, very fast jobs at a large scale.

/to @Tapjoy/dmmeas

### DIFF
--- a/lib/chore/hooks.rb
+++ b/lib/chore/hooks.rb
@@ -15,7 +15,8 @@ module Chore
 
   private
     def hooks_for(event)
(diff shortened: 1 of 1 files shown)


dankleiman:
Interesting that we hadn't hit the need for this memoization earlier, but we were close a couple years ago! MSTM

gregorysabatino:
LGTM

Issue #1, opened 2014-10-08
recommend a major version bump


deanxorr:
Chore `1.5.2` does not seem backwards compatible with earlier versions because the signature of the block passed into #handle_messages has changed.

StabbyCutyou:
This has already been resolved

PR #43, opened 2017-09-01
Various filesystem consumer improvements

obrie:
This makes a variety of performance / memory improvements to the filesystem consumer.  These are particularly effective when working with large backlogs.  This PR is best reviewed commit-by-commit.  Each change stands on its own, but I've bundled them into a single PR for the purpose of full context and how they're related.

Note that the reason I'm making these changes instead of investing time into the Go sidecar / FS queue is because there are only a few things we need to do to get our existing Ruby sidecar / FS queue behaving well enough to stand on its own.  Investing time into a Go implementation is a much larger chunk of work.

The following changes were made:

1. Only sleep between lookups in the filesystem consumer if files aren't found

Currently we sleep for 5s every time we we look at the filesystem for new files.  This means that we lose up to 5s of processing if our batch size is 100 and 100 new jobs appear immediately after we've consumed from the filesystem.

With this change, we will now only sleep if no new files have appeared since the last time we looked at the filesystem.  That is, if we look at the filesystem and there are no jobs to consume, then we'll sleep.

This helps avoid those times when we're just sitting there doing nothing even though there are jobs on the queue.

2. Determine fileystem consumer sleep time based on the consumer_sleep_interval Chore configuration

After the above change, we still now sleep for 5s when we don't find any new jobs on the filesystem which is still inefficient if the job workload is spiky.  With this change, we could drop the sleep time to 100ms via the consumer_sleep_interval configuration.  Even at 100ms, it's fast enough that we'll pick up new jobs pretty quickly off the filesystem and slow enough that we're not creating unnecessary CPU load.  The default will be 1s to match what we're doing in SQS.

3. Limit the number of files pulled from the filesystem on each iteration

Currently we pull all of the files off the filesystem at once when the consumer reads from the queue.  This can have a significant performance impact because if the queue is large enough we can exhaust our memory or open file limit.

Instead, we now only grab up to the configured queue_polling_size size at once.  With changes #1 and #2 in place, this is effectively as efficient as pulling all the files at once but without pulling it all into memory at once.

4. Don't keep files open forever (or at least until the file is deleted)

When we move job files from the new folder to the in progress folder, we open the file, lock it, but then never actually close the file.  Fortunately Ruby (or the OS) seems to be doing something smart here in that it somehow closes the open file when the file gets deleted.  However, it still leads to a lot more open files than there should be for the process.

With this change, we now close the file once we've locked it and moved it to "in progress".  Closing the file automatically unlocks it so we no longer need the unlock function.

5. Only lock files when moving them into the "in progress" directory

We only need to lock files when we're moving them from "new" to "in progress" in order to prevent us from moving files that are being written to by a publisher.  Once a consumer thread has moved it to "in progress", there's nothing any process is going to do to those files.  As a result, there's no need to lock the file again when moving it back to "new" since the consumer isn't competing with any other thread / process.

6. Don't block on files in the process of being written by the filesystem publisher

Currently we using a blocking lock when moving files from "new" to "in progress".  This is inefficient because if there are 1,000 jobs available to process and 1 job still being written by a publisher, we're going to wait on that publisher to finish writing even though we have 1,000 other jobs we could be processing.

Instead of using a blocking wait, this change moves us to a *non-blocking* wait.  It allows us to skip files that are still being written so that we can process all of the files that are ready.  On the next iteration of the consumer thread, we'll pick up the file that previously got skipped

7. Use Dir.glob instead of Dir.foreach in the filesystem consumer

This is more of a refactor and a forward-looking change than anything.  Currently we use `Dir.foreach` which pulls in directories like `.` and `..` in the results so we need to explicitly skip them.  Instead, we can use `Dir.glob` and avoid the need to check for directory names like that.

8. Support running multiple threaded consumers with the filesystem queue

Currently we only support a single consumer thread per queue.  This doesn't allow flexibility in the Chore configuration should we need to support multiple consumer threads (if, for example, we're bound by a single cpu for reading from the filesystem).

The change is actually fairly straightforward and, in fact, simplifies the filesystem queue configuration.  The main change needed here was just to be able to handle the fact that a file that another thread may have already moved into "in progress" could no longer exist.  Other than that, all of the existing functionality already supports parallel threads / processes reading job files from the "new" directory.

This also negates our need for the use of mutexes since the different consumer threads will handle the case when another thread is reading the same new job file.

9. Use File.delete instead of FileUtils.rm_f to reduce CPU usage in the filesystem consumer

`FileUtils.rm_f` does a lot of stuff under the hood that's necessary when all we want to do is delete a file off the filesystem.  This results in a bunch of additional CPU usage during high throughput scenarios.  By switching to the simpler `File.delete`, we effectively get the same functionality at a much lower cost to the CPU.

10. Support queue-specific configurations (like timeouts) in the filesystem consumer

Currently, the visibility timeout for every filesystem queue is based off the default queue timeout for the entire application.  This doesn't align with the behavior we have with SQS queues where the visibility timeout is configured at the source (i.e. on SQS itself).  In order to improve parity between the filesystem queue and SQS, this allows you to define queue-specific configurations on the filesystem.

Configurations will be read from a new "config" directory under the queue's root
directory (e.g. queue_name/config/timeout).  Currently the only configuration supported is `timeout`.

I felt that having individual files for different types of configurations would allow different extensions of chore to be able to have their own customized configurations.  We could do something like a JSON / YAML file, but it didn't seem necessary to have something that complicated just yet.  We could look at that in the future if we found the need.

## Follow-ups

There will be one more follow-up improvement (in a separate PR) to support multi-master mode with the filesystem queue.

## TODO

* [x] Verify changes locally

### DIFF
--- a/lib/chore.rb
+++ b/lib/chore.rb
@@ -41,6 +41,7 @@ module Chore #:nodoc:
     :shutdown_timeout      => (2 * 60),
     :max_attempts          => 1.0 / 0.0, # Infinity
     :dupe_on_cache_failure => false,
+    :queue_polling_size    => 10,
     :payload_handler       => Chore::Job,
     :master_procline       => "chore-master-#{Chore::VERSION}",
     :worker_procline       => "chore-worker-#{Chore::VERSION}",
--- a/lib/chore.rb
+++ b/lib/chore.rb
@@ -43,7 +43,8 @@ module Chore #:nodoc:
     :dupe_on_cache_failure => false,
     :payload_handler       => Chore::Job,
     :master_procline       => "chore-master-#{Chore::VERSION}",
-    :worker_procline       => "chore-worker-#{Chore::VERSION}"
+    :worker_procline       => "chore-worker-#{Chore::VERSION}",
+    :consumer_sleep_interval => 1
   }
 
   class << self
--- a/lib/chore/cli.rb
+++ b/lib/chore/cli.rb
@@ -144,6 +144,8 @@ module Chore #:nodoc:
       register_option 'shutdown_timeout', '--shutdown-timeout SECONDS', Float, "Upon shutdown, the number of seconds to wait before force killing worker strategies (default: #{Chore::DEFAULT_OPTIONS[:shutdown_timeout]})"
 
       register_option 'dupe_on_cache_failure', '--dupe-on-cache-failure BOOLEAN', 'Determines the deduping behavior when a cache connection error occurs. When set to false, the message is assumed not to be a duplicate. (default: false)'
+
+      register_option 'queue_polling_size', '--queue_polling_size NUM', Integer, 'Amount of messages to grab on each request (default: 10)'
     end
 
     def parse_opts(argv, ignore_errors = false) #:nodoc:
--- a/lib/chore/cli.rb
+++ b/lib/chore/cli.rb
@@ -135,7 +135,7 @@ module Chore #:nodoc:
         options[:consumer_strategy] = constantize(arg)
       end
 
-      register_option 'consumer_sleep_interval', '--consumer-sleep-interval INTERVAL', Float, 'Length of time in seconds to sleep when the consumer does not find any messages. Defaults vary depending on consumer implementation'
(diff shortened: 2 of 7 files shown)


obrie:
FYI this has been updated with a few fixes, additional improvements, and updated PR description.

jjrussell:
Looks really good. Nice explanation.

gregorysabatino:
LGTM

PR #44, opened 2017-09-01
Allow the master / worker proclines to be configured by the app

obrie:
We have some scenarios where we run multiple applications deployed onto the same servers each with their own master chore process.  Currently it's not possible to easily differentiate between these chore processes since they're all named the same.  This configuration would give applications the power to control the procline so that we can have different proclines and more easily monitor the processes.

/to @gregorysabatino @dankleiman @geoffreyclark


gregorysabatino:
LGTM

PR #45, opened 2017-09-01
Support multiple chore processes consuming from the filesystem queue

obrie:
In order to give us more flexibility in how we configure and deploy chore as a filesystem sidecar queue, this adds support for being able to run multiple chore masters or multiple forked workers.  Currently this is supported for external services like SQS; however, we don't have graceful support for it in the filesystem.

Similar to https://github.com/Tapjoy/chore/pull/43, this gets us a lot more flexibility in deployment configuration without investing time into a Go-based filesystem queue sidecar.

## Design

Currently the filesystem queue already supports multiple consumers across different processes.  However, the problem we still need to solve is what happens to in-progress jobs when a forked worker or master process dies.  The current implementation assumes only 1 master is ever running at a time.  This is because, on startup, it will immediately move everything out of "in progress" and into "new".  This works in a single master scenario since you can assume no one is processing the stuff that's in progress.  However, in multi-master, some of those jobs may be expired / leftover from old killed processes and some may be actively getting processed by a live master.

To address this, we take a page out of SQS's book.  We'll only ever delete from "in progress" if the job was completed *or* the job has expired based on the queue's visibility timeout.  The former is the normal scenario.  In the latter, jobs will expire if either (a) they took too long to process or (b) the process that owned the job was killed.

To track expiration time, we put a timestamp into the job's filename for when it was moved into "in progress".  This is necessary since neither the creation, modification, or access times change when the file moves.  This expiration time gets removed when the file is moved back into "new".

In addition to this periodic cleanup process, additionally error handling is needed in order to account for scenarios where we are both completing a job and cleaning up the job simulatenously.

## Performance

From a performance perspective, this is negligible unless there is a large backlog of files in "new".  In that case, we only allow the "new" folder to be cleaned up at most once a second in order to avoid a performance impact.  Once nice thing from this is that it actually allows for some delay before the job gets retried (jobs only get retried once every `visibility timeout` seconds).

## Summary

In summary, we've made the filesystem queue behave much more in line with how SQS behaves -- allowing us to make more sense of how jobs will behave regardless of the queue selected.

## TODO

* [x] Verify locally

### DIFF
--- a/lib/chore/consumer.rb
+++ b/lib/chore/consumer.rb
@@ -21,11 +21,6 @@ module Chore
     def self.reset_connection!
     end
 
-    # Cleans up any resources that were left behind from prior instances of the
-    # chore process.  By default, this is a no-op.
-    def self.cleanup(queue)
-    end
-
     # Consume takes a block with an arity of two. The two params are
     # |message_id,message_body| where message_id is any object that the
     # consumer will need to be able to act on a message later (reject, complete, etc)
--- a/lib/chore/queues/filesystem/consumer.rb
+++ b/lib/chore/queues/filesystem/consumer.rb
@@ -137,12 +137,9 @@ module Chore
           end
         end
 
+        # Rejects the given message from the filesystem by +id+. Currently a noop
         def reject(id)
-          Chore.logger.debug "Rejecting: #{id}"
-          make_new_again(id)
-        rescue Errno::ENOENT
-          # The job took too long to complete, was deemed expired, and moved
-          # back into "new".  Ignore.
+
         end
 
         def complete(id)
--- a/lib/chore/queues/filesystem/consumer.rb
+++ b/lib/chore/queues/filesystem/consumer.rb
@@ -22,22 +22,32 @@ module Chore
         Chore::CLI.register_option 'fs_queue_root', '--fs-queue-root DIRECTORY', 'Root directory for fs based queue'
         
         class << self
-          # Cleans up the in-progress files by making them new again.  This should only
-          # happen once per process.
-          def cleanup(queue)
-            new_dir = self.new_dir(queue)
-            in_progress_dir = self.in_progress_dir(queue)
+          # Cleans up expired in-progress files by making them new again.
+          def cleanup(expiration_time, new_dir, in_progress_dir)
+            each_file(File.join(in_progress_dir, '*.job')) do |job_file|
+              id, previous_attempts, timestamp = file_info(job_file)
+              next if timestamp > expiration_time
 
-            each_file(File.join(in_progress_dir, '*.job')) do |file|
-              make_new_again(file, new_dir, in_progress_dir)
+              begin
+                make_new_again(job_file, new_dir, in_progress_dir)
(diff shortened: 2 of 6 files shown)


obrie:
/to @jjrussell @dankleiman @gregorysabatino @geoffreyclark for early review

lib/chore/queues/filesystem/consumer.rb
```diff
@@ -99,7 +106,9 @@ def consume
           Chore.logger.info "Starting consuming file system queue #{@queue_name} in #{self.class.queue_dir(queue_name)}"
           while running?
             begin
-              #TODO move expired job files to new directory?
+              # Move expired job files to new directory?
```
dankleiman: nit: remove `?` in comment

obrie:
I've updated the PR description.  This should be ready for full review.

dankleiman:
LGTM!

lib/chore/consumer.rb
```diff
@@ -21,11 +21,6 @@ def initialize(queue_name, opts={})
     def self.reset_connection!
     end
 
-    # Cleans up any resources that were left behind from prior instances of the
-    # chore process.  By default, this is a no-op.
-    def self.cleanup(queue)
-    end
-
```
gregorysabatino: just curious, why are you removing this from the "base class"?
i get that we are now providing an implementation in filesystem consumer, but question maybe still relevant?

gregorysabatino: oh, think i got it now, nevermind

lib/chore/queues/filesystem/consumer.rb
```diff
@@ -99,7 +117,13 @@ def consume
           Chore.logger.info "Starting consuming file system queue #{@queue_name} in #{self.class.queue_dir(queue_name)}"
           while running?
             begin
-              #TODO move expired job files to new directory?
+              # Move expired job files to new directory (so long as enough time has
+              # passed since we last did this check)
+              if !@last_cleaned_at || (Time.now - @last_cleaned_at).to_i >= MINIMUM_EXPIRATION_CHECK_INTERVAL
```
gregorysabatino: was it intended to have this `MINIMUM_` get overridden?  just an assumption i made based on naming of the constant.  not something i view as a blocker fer sher.

gregorysabatino:
LGTM ; just 1 non-blocking question

lib/chore/queues/filesystem/consumer.rb
obrie: Good call, I'll rename this variable.  I could imagine some need for this to be configurable, but until that need is actually there, I would just leave this a constant.

obrie: Actually, I think the name may be more accurate as-is (I'll spend a few minutes thinking aout it).  The code doesn't guarantee 1s intervals -- it just guarantees that there's at least 1s between checks.

obrie: Thought about it more and decided to rename it, but leave in the comments indicating that this is more of a minimum value

PR #46, opened 2017-09-07
Use Dir.foreach instead of Dir.glob to fix a performance regression

obrie:
When a queue has a large backlog, `Dir.glob` performs much more poorly than `Dir.foreach`.  I had originally switched to `glob` to simplify the code, but the performance regression is way more than its worth.  See below for a directory with 200k jobs:

```
2.3.0 :015 > Benchmark.measure { index = 0; Dir.glob('/mnt/chore_fs_queue/development_web_requests/new/*.job') {|f| index += 1; break if index >= 1000 } }
 => #<Benchmark::Tms:0x0000000214cb50 @label="", @real=0.5653504389993032, @cstime=0.0, @cutime=0.0, @stime=0.2999999999999998, @utime=0.26000000000000023, @total=0.56> 
2.3.0 :016 > Benchmark.measure { index = 0; Dir.foreach('/mnt/chore_fs_queue/development_web_requests/new/') {|f| index += 1; break if index >= 1000 } }
 => #<Benchmark::Tms:0x000000031ef318 @label="", @real=0.0012892550003016368, @cstime=0.0, @cutime=0.0, @stime=0.0, @utime=0.0, @total=0.0> 
```

As you can see, `Dir.foreach` can be over 500x faster.  This is pretty critical so that the majority of your time isn't spent just listing files off the filesystem.

/to @gregorysabatino @dankleiman @geoffreyclark

### DIFF
--- a/lib/chore/queues/filesystem/consumer.rb
+++ b/lib/chore/queues/filesystem/consumer.rb
@@ -24,7 +24,7 @@ module Chore
         class << self
           # Cleans up expired in-progress files by making them new again.
           def cleanup(expiration_time, new_dir, in_progress_dir)
-            each_file(File.join(in_progress_dir, '*.job')) do |job_file|
+            each_file(in_progress_dir) do |job_file|
               id, previous_attempts, timestamp = file_info(job_file)
               next if timestamp > expiration_time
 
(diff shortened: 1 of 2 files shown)


obrie:
You'll note that this is basically a revert of fix/fs_performance_regression

PR #47, opened 2017-09-07
Fix the filesystem consumer processing incomplete job files

obrie:
When the publisher writes jobs to the filesystem, there's a small window of time where the file is on the filesystem with no data and can be locked by the consumer.  To avoid reading those files, the consumer will now only process files that actually have data in them.

With this change, it's now also possible that the publisher can create a file on the filesystem and then subsequently die, leaving a zero-length file that never gets processed by the consumer.

To avoid leaving empty files on the filesystem forever, the consumer will check for zero-length files that have exceeded the queue's visibility timeout and automatically delete them.

I started to see this error more frequently when I reduced the consumer read interval from 5s to 100ms.  This gave the consumer a much higher chance of actually grabbing the file off the filesystem before the publisher had completed writing to it.

/to @jjrussell @gregorysabatino @dankleiman @geoffreyclark

### DIFF
--- a/lib/chore/queues/filesystem/consumer.rb
+++ b/lib/chore/queues/filesystem/consumer.rb
@@ -42,21 +42,38 @@ module Chore
 
           # Moves job file to inprogress directory and returns the full path
           # if the job was successfully locked by this consumer
-          def make_in_progress(job, new_dir, in_progress_dir)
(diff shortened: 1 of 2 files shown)


dankleiman:
MSTM with timing change

PR #48, opened 2017-09-13
Fix filesystem publisher potentially leaving files locked until the Ruby process exits

obrie:
This makes a few fixes to how we publish files to the filesystem from chore:

1. When the filesystem publisher chooses a filename, it's possible that it conflicts with another file that already exists if the file gets created within a short enough timeframe in the same process.  When this happens, the file will be left in a permanently locked state until the Ruby process exits since we don't explicitly unlock the file in that case.

2. The file mutex isn't used because we only publish one file at a time in a single process (not multi-threaded) and we already have handling for conflicting files.

3. By using "w" as the file flag, Ruby documents that this can truncate existing files.  To avoid that, I've switched it to "a" which will either create or append files.  In the append case, we'll never actually append because we only write to the file if it's size is 0.

4. Add the process id into the filename for the job to reduce the potential for filename conflicts between 2 separate publisher processes.

/to @gregorysabatino @dankleiman @jjrussell

### DIFF
--- a/lib/chore/queues/filesystem/publisher.rb
+++ b/lib/chore/queues/filesystem/publisher.rb
@@ -33,7 +33,8 @@ module Chore
         def filename(queue_name, job_name)
           now = Time.now.strftime "%Y%m%d-%H%M%S-%6N"
           previous_attempts = 0
-          File.join(new_dir(queue_name), "#{queue_name}-#{job_name}-#{now}.#{previous_attempts}.job")
+          pid = Process.pid
+          File.join(new_dir(queue_name), "#{queue_name}-#{job_name}-#{pid}-#{now}.#{previous_attempts}.job")
         end
       end
     end
--- a/lib/chore/queues/filesystem/publisher.rb
+++ b/lib/chore/queues/filesystem/publisher.rb
@@ -20,7 +20,7 @@ module Chore
           while !published
             # keep trying to get a file with nothing in it meaning we just created it
             # as opposed to us getting someone else's file that hasn't been processed yet.
-            File.open(filename(queue_name, job[:class].to_s), "w") do |f|
+            File.open(filename(queue_name, job[:class].to_s), "a") do |f|
               if f.flock(File::LOCK_EX | File::LOCK_NB) && f.size == 0
                 f.write(encoded_job)
                 published = true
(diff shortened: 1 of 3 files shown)


lib/chore/queues/filesystem/publisher.rb
```diff
@@ -10,27 +10,20 @@ class Publisher < Chore::Publisher
...
           # First try encoding the job to avoid writing empty job files if this fails
           encoded_job = encode_job(job)
 
-          FILE_MUTEX.synchronize do
-            while true
-              # keep trying to get a file with nothing in it meaning we just created it
-              # as opposed to us getting someone else's file that hasn't been processed yet.
-              f = File.open(filename(queue_name, job[:class].to_s), "w")
+          published = false
+          while !published
```
gregorysabatino: i'm sure you thought of this, but i am not intimately familiar with chore, so ...
what are the chances we get stuck in this loop forever?

obrie: Well, consider how `filename` works.  On each iteration, it's using a combination of the process id and timestamp (to 6 decimal places) to generate a filename.  I can't think of a situation where this loop would infinitely keep generating filenames that conflict with filenames already on the filesystem -- unless time was frozen :)

gregorysabatino: but what about the impact of climate change?
clearly didn't trace the code path :)
LGTM

lib/chore/queues/filesystem/publisher.rb
```diff
@@ -40,7 +33,8 @@ def publish(queue_name,job)
         def filename(queue_name, job_name)
           now = Time.now.strftime "%Y%m%d-%H%M%S-%6N"
           previous_attempts = 0
-          File.join(new_dir(queue_name), "#{queue_name}-#{job_name}-#{now}.#{previous_attempts}.job")
+          pid = Process.pid
+          File.join(new_dir(queue_name), "#{queue_name}-#{job_name}-#{pid}-#{now}.#{previous_attempts}.job")
```
ehealy: Non blocking nit: I'm not currently seeing where we increment `previous_attempts` anywhere, so does it make sense to have it in the filename?

obrie: It gets incremented by the consumer when there's an error.  See https://github.com/Tapjoy/chore/blob/master/lib/chore/queues/filesystem/consumer.rb#L88

ehealy: 👍  carry on.

PR #50, opened 2019-05-28
[DMMEAS-1646] adds chore queue option to dedupe on arbitrary value per job

adnissen:
The goal of this PR is to allow chore jobs to define a custom lambda to be called before `perform`, the results of which we would use as a key to dedupe the job runs. We currently do part of this today for sqs message ids (sqs is "at-least once") in `sqs/consumer.rb`. Part of this PR refactors the code so that we are either doing one dedupe: using the custom job-defined function, or falling back to using the message id. 

A random sided-benefit of this PR is that by moving the deduping code from the `sqs/consumer` class to the `worker`, we actually _gain_ the ability to dedupe on filesystem queues which wasn't previously possible before.

One thing to keep in mind when reviewing this code is that the same chore triggers should be called as they were previously on duplicate messages. The behavior for dedupe-key-duplicates is slightly different, but making sure call the same callbacks is important.

I've also added some specs and have done manual testing by doing the following:
1. Modifying a tapinabox so that a job class has a `dedupe_lambda` defined, like:
```
class NonManagedSendCurrencyJob
  include Chore::Job
  include SendCurrencyJobHelper

  queue_options :name => "v2_NonManagedSendCurrency", :dedupe_lambda => lambda {|reward_id, options = {}| reward_id }
end
```
2. Generate a uuid in the console and call the job several times very quickly.
```
2.4.1 :006 > u = UUIDTools::UUID.random_create.to_s
 => "38df22f9-1cec-4c8d-87b1-28fa23e273c6"
2.4.1 :007 > NonManagedSendCurrencyJob.perform_async(u)
^[[A[2019-06-03 15:36:48 +0000 (4124)] DEBUG : [AWS SQS 200 0.264393 0 retries] get_queue_url(:queue_name=>"tapinabox_andrew_v2_NonManagedSendCurrency")


[2019-06-03 15:36:48 +0000 (4124)] DEBUG : [AWS SQS 200 0.069394 0 retries] send_message(:message_body=>"{\"class\":\"NonManagedSendCurrencyJob\",\"args\":[\"38df22f9-1cec-4c8d-87b1-28fa23e273c6\"]}",:queue_url=>"https://sqs.us-east-1.amazonaws.com/331510376354/tapinabox_andrew_v2_NonManagedSendCurrency")

 => [[]]
2.4.1 :008 > NonManagedSendCurrencyJob.perform_async(u)
[2019-06-03 15:36:48 +0000 (4124)] DEBUG : [AWS SQS 200 0.064696 0 retries] get_queue_url(:queue_name=>"tapinabox_andrew_v2_NonManagedSendCurrency")

[2019-06-03 15:36:48 +0000 (4124)] DEBUG : [AWS SQS 200 0.107138 0 retries] send_message(:message_body=>"{\"class\":\"NonManagedSendCurrencyJob\",\"args\":[\"38df22f9-1cec-4c8d-87b1-28fa23e273c6\"]}",:queue_url=>"https://sqs.us-east-1.amazonaws.com/331510376354/tapinabox_andrew_v2_NonManagedSendCurrency")

 => [[]]
2.4.1 :009 > NonManagedSendCurrencyJob.perform_async(u)
[2019-06-03 15:36:48 +0000 (4124)] DEBUG : [AWS SQS 200 0.064533 0 retries] get_queue_url(:queue_name=>"tapinabox_andrew_v2_NonManagedSendCurrency")

[2019-06-03 15:36:49 +0000 (4124)] DEBUG : [AWS SQS 200 0.068809 0 retries] send_message(:message_body=>"{\"class\":\"NonManagedSendCurrencyJob\",\"args\":[\"38df22f9-1cec-4c8d-87b1-28fa23e273c6\"]}",:queue_url=>"https://sqs.us-east-1.amazonaws.com/331510376354/tapinabox_andrew_v2_NonManagedSendCurrency")

 => [[]]
2.4.1 :010 > NonManagedSendCurrencyJob.perform_async(u)
[2019-06-03 15:36:49 +0000 (4124)] DEBUG : [AWS SQS 200 0.06444 0 retries] get_queue_url(:queue_name=>"tapinabox_andrew_v2_NonManagedSendCurrency")

[2019-06-03 15:36:49 +0000 (4124)] DEBUG : [AWS SQS 200 0.071577 0 retries] send_message(:message_body=>"{\"class\":\"NonManagedSendCurrencyJob\",\"args\":[\"38df22f9-1cec-4c8d-87b1-28fa23e273c6\"]}",:queue_url=>"https://sqs.us-east-1.amazonaws.com/331510376354/tapinabox_andrew_v2_NonManagedSendCurrency")

 => [[]]
2.4.1 :011 > NonManagedSendCurrencyJob.perform_async(u)
[2019-06-03 15:36:49 +0000 (4124)] DEBUG : [AWS SQS 200 0.064222 0 retries] get_queue_url(:queue_name=>"tapinabox_andrew_v2_NonManagedSendCurrency")

[2019-06-03 15:36:49 +0000 (4124)] DEBUG : [AWS SQS 200 0.066898 0 retries] send_message(:message_body=>"{\"class\":\"NonManagedSendCurrencyJob\",\"args\":[\"38df22f9-1cec-4c8d-87b1-28fa23e273c6\"]}",:queue_url=>"https://sqs.us-east-1.amazonaws.com/331510376354/tapinabox_andrew_v2_NonManagedSendCurrency")

 => [[]]
2.4.1 :012 > NonManagedSendCurrencyJob.perform_async(u)
[2019-06-03 15:36:49 +0000 (4124)] DEBUG : [AWS SQS 200 0.065662 0 retries] get_queue_url(:queue_name=>"tapinabox_andrew_v2_NonManagedSendCurrency")

[2019-06-03 15:36:49 +0000 (4124)] DEBUG : [AWS SQS 200 0.067002 0 retries] send_message(:message_body=>"{\"class\":\"NonManagedSendCurrencyJob\",\"args\":[\"38df22f9-1cec-4c8d-87b1-28fa23e273c6\"]}",:queue_url=>"https://sqs.us-east-1.amazonaws.com/331510376354/tapinabox_andrew_v2_NonManagedSendCurrency")

 => [[]]
2.4.1 :013 > NonManagedSendCurrencyJob.perform_async(u)
[2019-06-03 15:36:50 +0000 (4124)] DEBUG : [AWS SQS 200 0.065321 0 retries] get_queue_url(:queue_name=>"tapinabox_andrew_v2_NonManagedSendCurrency")

[2019-06-03 15:36:50 +0000 (4124)] DEBUG : [AWS SQS 200 0.069942 0 retries] send_message(:message_body=>"{\"class\":\"NonManagedSendCurrencyJob\",\"args\":[\"38df22f9-1cec-4c8d-87b1-28fa23e273c6\"]}",:queue_url=>"https://sqs.us-east-1.amazonaws.com/331510376354/tapinabox_andrew_v2_NonManagedSendCurrency")

 => [[]]
```
3. Observe the chore logs and see that perform is only called *once* every 40 or so seconds (the `ERROR` lines), while the duplicates to come in after the first run are deleted.
```
Jun  5 15:30:29 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:30:28 +0000 (1786)] INFO : Running job NonManagedSendCurrencyJob with params {"class"=>"NonManagedSendCurrencyJob", "args"=>["09d3fd0a-5736-4b25-8535-1f63675f0a4b"]}
Jun  5 15:30:29 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:30:28 +0000 (1786)] ERROR : Failed to run job {"class":"NonManagedSendCurrencyJob","args":["09d3fd0a-5736-4b25-8535-1f63675f0a4b"]} with error: Reward not found: 09d3fd0a-5736-4b25-8535-1f63675f0a4b at /opt/apps/tapjoyserver/releases/tapjoyserver-20190605135114-98dea6193b/app/helpers/send_currency_job_helper.rb:52:in `perform'
Jun  5 15:30:29 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:30:28 +0000 (1786)] INFO : Found and deleted duplicate job NonManagedSendCurrencyJob
Jun  5 15:30:29 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:30:28 +0000 (1786)] INFO : Found and deleted duplicate job NonManagedSendCurrencyJob
Jun  5 15:30:29 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:30:29 +0000 (1786)] INFO : Found and deleted duplicate job NonManagedSendCurrencyJob
Jun  5 15:31:30 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:30:29 +0000 (1786)] INFO : Found and deleted duplicate job NonManagedSendCurrencyJob
Jun  5 15:31:30 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:30:29 +0000 (1786)] INFO : Found and deleted duplicate job NonManagedSendCurrencyJob
Jun  5 15:31:30 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:30:59 +0000 (1786)] INFO : Running job NonManagedSendCurrencyJob with params {"class"=>"NonManagedSendCurrencyJob", "args"=>["09d3fd0a-5736-4b25-8535-1f63675f0a4b"]}
Jun  5 15:31:30 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:30:59 +0000 (1786)] ERROR : Failed to run job {"class":"NonManagedSendCurrencyJob","args":["09d3fd0a-5736-4b25-8535-1f63675f0a4b"]} with error: Reward not found: 09d3fd0a-5736-4b25-8535-1f63675f0a4b at /opt/apps/tapjoyserver/releases/tapjoyserver-20190605135114-98dea6193b/app/helpers/send_currency_job_helper.rb:52:in `perform'
Jun  5 15:31:30 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:31:30 +0000 (1786)] INFO : Running job NonManagedSendCurrencyJob with params {"class"=>"NonManagedSendCurrencyJob", "args"=>["09d3fd0a-5736-4b25-8535-1f63675f0a4b"]}
Jun  5 15:31:30 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:31:30 +0000 (1786)] ERROR : Failed to run job {"class":"NonManagedSendCurrencyJob","args":["09d3fd0a-5736-4b25-8535-1f63675f0a4b"]} with error: Reward not found: 09d3fd0a-5736-4b25-8535-1f63675f0a4b at /opt/apps/tapjoyserver/releases/tapjoyserver-20190605135114-98dea6193b/app/helpers/send_currency_job_helper.rb:52:in `perform'
Jun  5 15:32:37 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:32:05 +0000 (1786)] INFO : Running job NonManagedSendCurrencyJob with params {"class"=>"NonManagedSendCurrencyJob", "args"=>["09d3fd0a-5736-4b25-8535-1f63675f0a4b"]}
Jun  5 15:32:37 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:32:05 +0000 (1786)] ERROR : Failed to run job {"class":"NonManagedSendCurrencyJob","args":["09d3fd0a-5736-4b25-8535-1f63675f0a4b"]} with error: Reward not found: 09d3fd0a-5736-4b25-8535-1f63675f0a4b at /opt/apps/tapjoyserver/releases/tapjoyserver-20190605135114-98dea6193b/app/helpers/send_currency_job_helper.rb:52:in `perform'
Jun  5 15:32:37 ip-10-189-97-151 tapjoyserver-jobs[1]: [2019-06-05 15:32:37 +0000 (1786)] INFO : Running job NonManagedSendCurrencyJob with params {"class"=>"NonManagedSendCurrencyJob", "args"=>["09d3fd0a-5736-4b25-8535-1f63675f0a4b"]}
```

### DIFF
--- a/lib/chore/job.rb
+++ b/lib/chore/job.rb
@@ -60,6 +60,12 @@ module Chore
             raise ArgumentError, "#{self.to_s}: backoff must accept a single argument"
           end
         end
+
+        if @chore_options.key?(:dedupe_lambda)
+          if !@chore_options[:dedupe_lambda].is_a?(Proc)
+            raise ArgumentError, "#{self.to_s}: dedupe_lambda must be a lambda or Proc"
+          end
+        end
       end
 
       # This is a method so it can be overriden to create additional required
@@ -107,6 +113,17 @@ module Chore
       # the key at this point.
       def has_backoff?
         self.options.key?(:backoff)
+      end
+
+      def has_dedupe_lambda?
+        self.options.key?(:dedupe_lambda)
+      end
+
+      def dedupe_key(*args)
+        return unless has_dedupe_lambda?
+
+        # run the proc to get the key
+        self.options[:dedupe_lambda].call(*args).to_s
       end
     end #ClassMethods
 
--- a/lib/chore/worker.rb
+++ b/lib/chore/worker.rb
@@ -42,6 +42,20 @@ module Chore
       @started_at + total_timeout
     end
 
+    def duplicate_work?(item)
+      # if we've got a duplicate, remove the message from the queue by not actually running and also not reporting any errors
+      payload = options[:payload_handler].payload(item.decoded_message)
+
+       # if we're hitting the custom dedupe key, we want to remove this message from the queue
+      if item.klass.has_dedupe_lambda? && item.consumer.duplicate_message?(item.klass.dedupe_key(*payload), item.klass, item.queue_timeout)
+        Chore.logger.info { "Found and deleted duplicate job #{item.klass}"}
+        item.consumer.complete(item.id)
+        return true
+      end
+
+      return false
+    end
+
     # The workhorse. Do the work, all of it. This will block for an entirely unspecified amount
     # of time based on the work to be performed. This will:
     # * Decode each message.
@@ -58,6 +72,9 @@ module Chore
         begin
           item.decoded_message = options[:payload_handler].decode(item.message)
           item.klass = options[:payload_handler].payload_class(item.decoded_message)
+
+          next if duplicate_work?(item)
+
           Chore.run_hooks_for(:worker_to_start, item)
           start_item(item)
         rescue => e
--- a/spec/chore/worker_spec.rb
+++ b/spec/chore/worker_spec.rb
@@ -2,6 +2,10 @@ require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
 
 describe Chore::Worker do
 
+  before(:each) do
+    allow(consumer).to receive(:duplicate_message?).and_return(false)
+  end
+
   class SimpleJob
     include Chore::Job
     queue_options :name => 'test',
@@ -10,6 +14,30 @@ describe Chore::Worker do
 
     def perform(*args)
       return args
+    end
+  end
+
+  class SimpleDedupeJob
+    include Chore::Job
+    queue_options :name => 'dedupe_test',
+      :publisher => FakePublisher,
+      :max_attempts => 100,
+      :dedupe_lambda => lambda { |first, second, third| first }
+    
+    def perform(first, second, third)
+      return second
+    end
+  end
+
+  class InvalidDedupeJob
+    include Chore::Job
+    queue_options :name => 'invalid_dedupe_test',
+      :publisher => FakePublisher,
+      :max_attempts => 100,
+      :dedupe_lambda => lambda { |first, second, third| first }
+    
+    def perform(first, second)
+      return second
     end
   end
 
@@ -39,6 +67,33 @@ describe Chore::Worker do
       SimpleJob.should_receive(:perform).exactly(10).times
       consumer.should_receive(:complete).exactly(10).times
       Chore::Worker.start(work, {:payload_handler => payload_handler})
+    end
+
+    context 'when the job has a dedupe_lambda defined' do
+      context 'when the value being deduped on is unique' do
+        let(:job_args) { [rand,2,'3'] }
+        let(:encoded_job) { Chore::Encoder::JsonEncoder.encode(job) }
+        let(:job) { SimpleDedupeJob.job_hash(job_args) }
+        it 'should call complete for each unique value' do
+          allow(consumer).to receive(:duplicate_message?).and_return(false)
+          work = []
+          work << Chore::UnitOfWork.new(1, 'dedupe_test', 60, Chore::Encoder::JsonEncoder.encode(SimpleDedupeJob.job_hash([rand,2,'3'])), 0, consumer)
+          SimpleDedupeJob.should_receive(:perform).exactly(1).times
+          consumer.should_receive(:complete).exactly(1).times
+          Chore::Worker.start(work, {:payload_handler => payload_handler})
+        end
+      end
+
+      context 'when the dedupe lambda does not take the same number of arguments as perform' do
+        it 'should raise an error and not complete the job' do
+          work = []
+          work << Chore::UnitOfWork.new(1, 'invalid_dedupe_test', 60, Chore::Encoder::JsonEncoder.encode(InvalidDedupeJob.job_hash([rand,2,'3'])), 0, consumer)
+          work << Chore::UnitOfWork.new(2, 'invalid_dedupe_test', 60, Chore::Encoder::JsonEncoder.encode(InvalidDedupeJob.job_hash([rand,2,'3'])), 0, consumer)
+          work << Chore::UnitOfWork.new(1, 'invalid_dedupe_test', 60, Chore::Encoder::JsonEncoder.encode(InvalidDedupeJob.job_hash([rand,2,'3'])), 0, consumer)
+          consumer.should_not_receive(:complete)
+          Chore::Worker.start(work, {:payload_handler => payload_handler})
+        end
+      end
     end
   end
 
--- a/lib/chore/consumer.rb
+++ b/lib/chore/consumer.rb
@@ -53,5 +53,15 @@ module Chore
     def provide_work(n)
       raise NotImplementedError
     end
+
+    # now, given an arbitrary key and klass, have we seen the key already?
+    def duplicate_message?(dedupe_key, klass, queue_timeout)
+      dupe_detector.found_duplicate?(:id=>dedupe_key, :queue=>klass.to_s, :visibility_timeout=>queue_timeout)
+    end
+
+    def dupe_detector
+      @dupes ||= DuplicateDetector.new({:servers => Chore.config.dedupe_servers,
+                                        :dupe_on_cache_failure => false})
+    end
   end
 end
--- a/lib/chore/queues/sqs/consumer.rb
+++ b/lib/chore/queues/sqs/consumer.rb
@@ -70,24 +70,12 @@ module Chore
           msg = queue.receive_messages(:limit => sqs_polling_amount, :attributes => [:receive_count])
           messages = *msg
           messages.each do |message|
-            unless duplicate_message?(message)
-              block.call(message.handle, queue_name, queue_timeout, message.body, message.receive_count - 1)
+            unless duplicate_message?(message.id, message.queue.url, message.queue.visibility_timeout)
+              block.call(message.handle, queue_name, message.queue.visibility_timeout, message.body, message.receive_count - 1)
             end
             Chore.run_hooks_for(:on_fetch, message.handle, message.body)
           end
           messages
-        end
-
-        # Checks if the given message has already been received within the timeout window for this queue
-        def duplicate_message?(message)
-          dupe_detector.found_duplicate?(:id=>message.id, :queue=>message.queue.url, :visibility_timeout=>message.queue.visibility_timeout)
-        end
-
-        # Returns the instance of the DuplicateDetector used to ensure unique messages.
-        # Will create one if one doesn't already exist
-        def dupe_detector
-          @dupes ||= DuplicateDetector.new({:servers => Chore.config.dedupe_servers,
-                                            :dupe_on_cache_failure => Chore.config.dupe_on_cache_failure})
         end
 
         # Retrieves the SQS queue with the given +name+. The method will cache the results to prevent round trips on
--- a/lib/chore/version.rb
+++ b/lib/chore/version.rb
@@ -1,7 +1,7 @@
 module Chore
   module Version #:nodoc:
     MAJOR = 3
-    MINOR = 1
+    MINOR = 2
     PATCH = 0
 
     STRING = [ MAJOR, MINOR, PATCH ].join('.')
--- a/spec/chore/duplicate_detector_spec.rb
+++ b/spec/chore/duplicate_detector_spec.rb
@@ -2,7 +2,21 @@ require 'spec_helper'
 require 'securerandom'
 
 describe Chore::DuplicateDetector do
-  let(:memcache) { double("memcache") }
+  class FakeDalli
+    def initialize
+      @store = {}
+    end
+    def add(id, val, ttl=0)
+      if @store[id] && @store[id][:inserted] + @store[id][:ttl] > Time.now.to_i
+        return false
+      else
+        @store[id] = {:val => val, :ttl => ttl, :inserted => Time.now.to_i}
+        return true
+      end
+    end
+  end
+
+  let(:memcache) { FakeDalli.new }
   let(:dupe_on_cache_failure) { false }
   let(:dedupe_params)  { { :memcached_client => memcache, :dupe_on_cache_failure => dupe_on_cache_failure } }
   let(:dedupe) { Chore::DuplicateDetector.new(dedupe_params)}
@@ -15,12 +29,11 @@ describe Chore::DuplicateDetector do
 
   describe "#found_duplicate" do
     it 'should not return true if the message has not already been seen' do
-      expect(memcache).to receive(:add).and_return(true)
       expect(dedupe.found_duplicate?(message_data)).to_not be true
     end
 
     it 'should return true if the message has already been seen' do
-      expect(memcache).to receive(:add).and_return(false)
+      memcache.add(message_data[:id], 1, message_data[:visibility_timeout])
       expect(dedupe.found_duplicate?(message_data)).to be true
     end
 
@@ -34,13 +47,12 @@ describe Chore::DuplicateDetector do
     end
 
     it "should set the timeout to be the queue's " do
-      expect(memcache).to receive(:add).with(id,"1",timeout).and_return(true)
+      expect(memcache).to receive(:add).with(id,"1",timeout).and_call_original
       expect(dedupe.found_duplicate?(message_data)).to be false
     end
 
     it "should call #visibility_timeout once and only once" do
       expect(queue).to receive(:visibility_timeout).once
-      expect(memcache).to receive(:add).at_least(3).times.and_return(true)
       3.times { dedupe.found_duplicate?(message_data) }
     end
 
--- a/spec/chore/strategies/worker/forked_worker_strategy_spec.rb
+++ b/spec/chore/strategies/worker/forked_worker_strategy_spec.rb
@@ -23,6 +23,10 @@ describe Chore::Strategy::ForkedWorkerStrategy do
   let!(:worker) { Chore::Worker.new(job) }
   let(:pid) { Random.rand(2048) }
 
+  before(:each) do
+    allow(consumer).to receive(:duplicate_message?).and_return(false)
+  end
+
   after(:each) do
     allow(Process).to receive(:kill) { nil }
     allow(Process).to receive(:wait) { pid }


lib/chore/worker.rb
```diff
@@ -86,7 +86,7 @@ def start_item(item)
 
       begin
         Chore.logger.info { "Running job #{klass} with params #{message}"}
-        perform_job(klass,message)
+        perform_job(klass, item.id, message, item.queue_timeout)
```
adnissen: I tried to pass just `item` here, but for reasons I TRULY DO NOT UNDERSTAND, whatever the first parameter passed into `perform_job` is `includes` the `job` class, which triggers https://github.com/Tapjoy/chore/pull/50/files#diff-41d7585164fb76e745b1878b6650d4ebR26, so the first parameter _has_ to be the `klass`. In for a penny, in for a pound, so I just decided to make the other parameters explicit as well.

If anyone can understand why this is happening I'd be very interested.

lib/chore/job.rb
```diff
@@ -60,6 +60,14 @@ def queue_options(opts = {})
             raise ArgumentError, "#{self.to_s}: backoff must accept a single argument"
           end
         end
+
+        if @chore_options.key?(:dedupe_lambda)
+          if !@chore_options[:dedupe_lambda].is_a?(Proc)
+            raise ArgumentError, "#{self.to_s}: dedupe_lambda must be a lambda"
+          elsif @chore_options[:dedupe_lambda].arity != self.new.method(:perform).arity
+            raise ArgumentError, "#{self.to_s}: dedupe_lambda must accept the same number of arguments as perform (perform must be defined _after_ queue_options in the job class)"
```
adnissen: bizarre requirement, is there a way to throw this later

lib/chore/worker.rb
```diff
@@ -123,8 +123,28 @@ def handle_failure(item, message, klass, e)
...
+    def perform_job(klass, id, message, queue_timeout)
+      payload = options[:payload_handler].payload(message)
+
+      # if they have a dedupe lambda, get their custom dedupe key
+      # we need to wrap this in a rescue block in case the user-defined lambda throws an error
+      id =
+        begin
+          klass.has_dedupe_lambda? ? klass.dedupe_key(*payload) : id
+        rescue => e
+          id
```
adnissen: report this exception

adnissen:
is calling all the hooks for duplicates a problem?

adnissen:
calling `complete` deletes all the messages from the queue with the same id. this is a problem if we have two messages with the same sqs id and the first fails (and thus is added to the back of the queue), the 2nd will 'complete' (because it is a duplicate) and then delete the first (uncompleted) job from the queue.

adnissen:
filesystem queue doesn't really need this, can we make it optional

adnissen:
don't connect to the local memcached, try fakememached

lib/chore/duplicate_detector.rb
```diff
@@ -23,7 +14,21 @@ def initialize(opts={})
       @dupe_on_cache_failure = opts.fetch(:dupe_on_cache_failure) { false }
       @timeout               = opts.fetch(:timeout) { 0 }
       @servers               = opts.fetch(:servers) { nil }
-      @memcached_client      = opts.fetch(:memcached_client) { Dalli::Client.new(@servers, memcached_options) }
+      @memcached_client      = opts.fetch(:memcached_client) { memcached_client(memcached_options) }
+    end
+
+    def memcached_client(options={})
```
nwmartin: So the idea is that we only complain if you actually attempt to make the call?

lib/chore/duplicate_detector.rb
```diff
@@ -23,7 +14,21 @@ def initialize(opts={})
       @dupe_on_cache_failure = opts.fetch(:dupe_on_cache_failure) { false }
       @timeout               = opts.fetch(:timeout) { 0 }
       @servers               = opts.fetch(:servers) { nil }
-      @memcached_client      = opts.fetch(:memcached_client) { Dalli::Client.new(@servers, memcached_options) }
+      @memcached_client      = opts.fetch(:memcached_client) { memcached_client(memcached_options) }
```
nwmartin: Strictly speaking this is already being set by making the call but it does make the code line up nice.

lib/chore/duplicate_detector.rb
adnissen: yea, I just need a way to override the `memcached_client` object _without_ overriding the `Dalli::Client` method (which would require `dalli` in the spec files).

lib/chore/job.rb
```diff
@@ -60,6 +60,12 @@ def queue_options(opts = {})
             raise ArgumentError, "#{self.to_s}: backoff must accept a single argument"
           end
         end
+
+        if @chore_options.key?(:dedupe_lambda)
+          if !@chore_options[:dedupe_lambda].is_a?(Proc)
+            raise ArgumentError, "#{self.to_s}: dedupe_lambda must be a lambda"
```
nwmartin: Strictly speaking lambdas are only a subset of Procs. I'm not sure that the difference will matter much here.

lib/chore/worker.rb
```diff
@@ -42,6 +42,24 @@ def expires_at
...
     end
 
+    def duplicate_work?(item)
+      # if we've got a duplicate, remove the message from the queue by not actually running and also not reporting any errors
+      payload = options[:payload_handler].payload(item.decoded_message)
+      
+      # only dedupe on the message id if we're getting work from an sqs consumer
+      return false if item.consumer.class != Chore::Queues::SQS::Consumer
+
+      # if the message is a duplicate on the queue id, re-enque it
```
nwmartin: `re-enqueue`

lib/chore/worker.rb
```diff
@@ -42,6 +42,24 @@ def expires_at
       @started_at + total_timeout
     end
 
+    def duplicate_work?(item)
+      # if we've got a duplicate, remove the message from the queue by not actually running and also not reporting any errors
+      payload = options[:payload_handler].payload(item.decoded_message)
+      
+      # only dedupe on the message id if we're getting work from an sqs consumer
```
nwmartin: I don't understand this comment, is this so we don't fall over between SQS and the file consumer?

lib/chore/job.rb
adnissen: yea, it seems like the only way to check if something is a lambda would be to do something like `@chore_options[:dedupe_lambda].respond_to? :call` which seems weirder. If they  pass a proc, it'll fail on the arity check anyways

lib/chore/worker.rb
adnissen: yea, there's really no way for a duplicate-by-queue-id filesystem message to come in, so there's no sense in storing all of those in the cache

adnissen: I'll update that comment

lib/chore/worker.rb
nwmartin: If we have a dupe on message id alone, why would we re-enqueue it? We'd want to silently kill this message right?

nwmartin: No, no we don't. Right. If the message ids are the same, we just want to ignore this one because we're expecting a real job run to get through at some point.

We only want to ack if they have DIFFERENT message ids.

lib/chore/worker.rb
```diff
@@ -42,6 +42,24 @@ def expires_at
...
+      return false if item.consumer.class != Chore::Queues::SQS::Consumer
+
+      # if the message is a duplicate on the queue id, re-enque it
+      return true if duplicate_message?(item.id, item.klass, item.queue_timeout)
+      
+       # if we're hitting the custom dedupe key, we want to remove this message from the queue
+      if item.klass.has_dedupe_lambda? && duplicate_message?(item.klass.dedupe_key(*payload), item.klass, item.queue_timeout)
+        Chore.logger.info { "Found and deleted duplicate job #{item.klass}"}
+        item.consumer.complete(item.id)
```
nwmartin: ...which we do right here in the case of custom keys. Sweet.

lib/chore/job.rb
nwmartin: Procs respond to call too.

lib/chore/worker.rb
```diff
@@ -42,6 +42,24 @@ def expires_at
       @started_at + total_timeout
     end
 
+    def duplicate_work?(item)
+      # if we've got a duplicate, remove the message from the queue by not actually running and also not reporting any errors
+      payload = options[:payload_handler].payload(item.decoded_message)
+      
+      # only dedupe on the message id if we're getting work from an sqs consumer (currently the only consumer that can recieve the same message id twice)
+      return false if item.consumer.class != Chore::Queues::SQS::Consumer
```
obrie: Could we base this off the presence of dedupe servers rather than the consumer class?

lib/chore/worker.rb
```diff
@@ -42,6 +42,24 @@ def expires_at
...
+
+      # if the message is a duplicate on the queue id, re-enqueue it
```
obrie: `re-enqueue it` isn't quite right here -- really we're assuming that some other worker has picked up the job and is processing it.  The job may or may not get re-enqueued depending on whether it succeeds on the other worker.

spec/chore/worker_spec.rb
```diff
@@ -2,6 +2,28 @@
...
+    end
+    def add(id, val, ttl)
+      if @store[id]
+        return false
+      else
+        @store[id] = val
+        Thread.new do
+          sleep ttl
+          @store.delete(id)
+        end
```
obrie: Using threads in specs like this can lead to unpredictable results / slow spaces.  In fakememcached, we store the ttl along with the value and check that when looking up the value.  I would suggest a similar approach here.

spec/chore/worker_spec.rb
```diff
@@ -2,6 +2,28 @@
...
+        return true
+      end
+    end
+  end
+
+  before(:each) do
+    allow_any_instance_of(Chore::DuplicateDetector).to receive(:memcached_client).and_return(FakeDalli.new)
```
obrie: Rather than stub the internal `memcached_client` method, we should be able to stub `Dalli::Client.new`.  That would give us the same behavior without having to worry about the internals of the duplicate detector.

lib/chore/job.rb
obrie: The arity only fails when you have default arguments in the proc.  I would suggest keeping the error messaging the same -- `backoff must be a lambda or Proc`.   Having some options take in a lambda-only and others take in lambda or proc is a little confusing.

lib/chore/job.rb
```diff
@@ -108,6 +114,25 @@ def prefixed_queue_name
...
+
+      def has_dedupe_lambda?
+        self.options.key?(:dedupe_lambda)
+      end
+
+      def dedupe_key(*args)
+        return unless has_dedupe_lambda?
+        if self.options[:dedupe_lambda].arity != self.instance_method(:perform).arity
+          raise ArgumentError, "#{self.to_s}: dedupe_lambda must accept the same number of arguments as perform"
+        end
```
obrie: At this point, I feel like this type of error checking is unnecessary.  You'll effectively get the same message when you execute `call` with a different set of arguments.  If someone feels like using a `Proc`, then they'll be able to do that.

lib/chore/job.rb
```diff
@@ -108,6 +114,25 @@ def prefixed_queue_name
...
+        # run the proc to get the key
+        begin
+          key = self.options[:dedupe_lambda].call(*args)
+        rescue => e
+          raise "#{self.to_s}: dedupe_lambda raised error: #{e}"
```
obrie: We lose stack trace and exception class name information when we re-raise like this.  This would end up reporting to NewRelic as StandardError, which could end up getting combined with other StandardErrors.  You also would only get a stack trace starting at this line.

I think we could get away without a begin/rescue here unless we actually want the error condition to change something about the behavior of the dedupe.

lib/chore/worker.rb
```diff
@@ -123,8 +144,19 @@ def handle_failure(item, message, klass, e)
       end
     end
 
-    def perform_job(klass, message)
-      klass.perform(*options[:payload_handler].payload(message))
+    def dupe_detector
+      @dupes ||= DuplicateDetector.new({:servers => Chore.config.dedupe_servers,
+                                        :dupe_on_cache_failure => false})
```
obrie: Instead of a single connection per consumer instance, we will now have a connection per Worker instance.  In some cases we have job servers with 30+ workers and just a single consumer.  Theoretically we could see more than an order of magnitude increase in the number of connections to the memcached cluster.  While our max is 65k, a connection per worker may actually have an impact on our cluster.  We may want to consider either talking to ops about this or leaving the majority of the dedupe work in the consumer and not the worker.

lib/chore/worker.rb
```diff
@@ -123,8 +144,19 @@ def handle_failure(item, message, klass, e)
       end
     end
 
-    def perform_job(klass, message)
-      klass.perform(*options[:payload_handler].payload(message))
+    def dupe_detector
+      @dupes ||= DuplicateDetector.new({:servers => Chore.config.dedupe_servers,
```
obrie: Now that `DuplicateDetector` is used in this class, we should add a `require 'chore/duplicate_detector'` at the top.  Currently it's in the `Consumer` class and should get moved out of there.

lib/chore/worker.rb
```diff
@@ -86,7 +103,7 @@ def start_item(item)
 
       begin
         Chore.logger.info { "Running job #{klass} with params #{message}"}
-        perform_job(klass,message)
+        perform_job(klass, item.id, message, item.queue_timeout)
```
adnissen: change this back

lib/chore/worker.rb
```diff
@@ -123,8 +140,10 @@ def handle_failure(item, message, klass, e)
       end
     end
 
-    def perform_job(klass, message)
-      klass.perform(*options[:payload_handler].payload(message))
+    def perform_job(klass, id, message, queue_timeout)
```
adnissen: change back

adnissen:
review @obrie @Tapjoy/dmmeas

adnissen:
I've built and deployed this as version 3.2.0

PR #54, opened 2020-08-20
SWAT-2340: Fix work being thrown away in thrashing preforked workers

obrie:
https://tapjoy.atlassian.net/browse/SWAT-2340

* Fix work being thrown away when it can't be assigned to a worker
* Fix dead workers being considered for work


obrie:
Apologies for the long-winded PR description, but it seemed complex enough to warrant it :smile:

PR #55, opened 2020-08-27
OPER-5336 Update AWS SDK library

alanbrent:
This PR updates the AWS SDK to a modern version with support for authentication via [Web Federated Identity](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/loading-browser-credentials-federated-id.html). While is designed to be a drop-in replacement for applications that already use Chore, there are various API changes in the AWS SDK, some of which have necessitated API changes in `Chore::Queues`. Hence I have incremented the major version of the gem.

Along the way I have made a bunch of QOL updates to the documentation. Hopefully people will find them useful :)

### Minor Changes

1. `Chore::Consumer#consume` signature has been unified between interface and implementations
1. `Chore::Consumer#handle_jobs` renamed to `Chore::Consumer::#handle_messages` to unify names between publisher and consumer interfaces
1. `Chore::Consumer#handle_messages` private method used in all consumers has been added to the interface definition
1. `Chore::Publisher#reset_connection!` added to interface definition
1. `Chore::Consumer`-s and `Chore::Publisher`-s unified to use `message_id` argument name rather than `id` where applicable

### Major Changes

1. `UnitOfWork` now takes `:receipt_handle` argument to accommodate AWS SDK requirements
    - this is passed in as `nil` in the filesystem handler, which doesn't need it
1. SQS client object is now allocated in the parent `Chore::Queues::SQS` module rather than being duplicated in the publisher and consumer classes
1. `Chore::Queues::Consumer#complete` signature now takes an options hash to accommodate changes needed by the SQS consumer
1. `Chore::Queues::SQS::Consumer#complete` now requires the SQS message receipt handle to be passed in the options hash
1. API interactions with SQS queues have been refactored to accommodate changes in the AWS SDK APIs

## Spec Changes

1. We're now sharing fake objects between all the SQS specs
1. We're now `double`-ing the actual AWS API objects rather than generically named doubles
    - Perhaps this is incorrect? I'm not well-versed on this stuff so definitely let me know!
1. Incorrect usage of `expect(thing).to receive(:action).and_return(expected_value)` has been fixed
   - This was confusing for me, and I learned that `and_return` is actually a _mock_ being _set_ rather than an expected _value_ to be _compared_ against
1. A couple of seemingly useless specs that weren't actually testing anything have been disabled and may yet be removed in this PR, pending feedback.

## Quality Of Life Changes

1. Many more YARD docs
1. Documented the release process

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,26 @@
 **Cleanups**
 - N/A
 
+## [v4.0.0](https://github.com/Tapjoy/chore/tree/v4.0.0)
+
+**Features**
+
+- AWS SDK library has been updated to support AWS authentication via
+[Web Federated Identity](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/loading-browser-credentials-federated-id.html)
+  - The API has jumped ahead 2 major versions, so some of the internals (e.g.
+    [`Chore::UnitOfWork`](lib/chore/unit_of_work.rb)) had to be changed to accommodate its changes. In the end, however,
+    this update is designed to function as a drop-in replacement for earlier versions of Chore
+
+**Fixed bugs**
+
+- Some of the SQS specs were not actually testing output values
+
+**Cleanups**
+
+- Many more YARD docs
+- Documented the release process
+- Mild overhaul of the base README
+
 ## [v3.1.0](https://github.com/Tapjoy/chore/tree/v3.1.0) (2017-09-15)
 
 **Features**
--- a/README.md
+++ b/README.md
@@ -1,92 +1,90 @@
-# Chore: Job processing... for the future!
+# Chore: Async Job Processing Framework For Ruby
 
 [![Build Status](https://travis-ci.org/Tapjoy/chore.svg?branch=master)](https://travis-ci.org/Tapjoy/chore)
 
 ## About
 
-Chore is a pluggable, multi-backend job processor. It was built from the ground up to be extremely flexible. We hope that you
-will find integrating and using Chore to be as pleasant as we do.
+Chore is a pluggable, multi-backend job processing framework. It was built from the ground up to be extremely flexible.
+We hope that you find integrating and using Chore to be as pleasant as we do.
 
-The full docs for Chore can always be found at http://tapjoy.github.io/chore.
+The full docs for Chore can always be found at https://tapjoy.github.io/chore.
 
 ## Configuration
 
 Chore can be integrated with any Ruby-based project by following these instructions:
 
-    gem 'chore-core', '~> 1.6.0'
+1. Add `chore-core` to the Gemfile
 
-If you also plan on using SQS, you must also bring in dalli to use for memcached:
+    ```ruby
+    gem 'chore-core', '~> 4.0.0'
+    ```
 
+    When using SQS, also add `dalli` to use for `memcached`-based deduplication:
+
+    ```ruby
     gem 'dalli'
+    ```
 
-Create a `Chorefile` file in the root of your project directory. While you can configure Chore itself from this file, it's primarly used to direct the Chore binary toward the root of your application, so that it can locate all of the depdendencies and required code.
+1. Create a `Chorefile` file in the root of the project directory. While Chore itself can be configured from this file,
+it's primarily used to direct the Chore binstub toward the root of the application so that it can locate all of the
+dependencies and required code.
 
+    ```
     --require=./<FILE_TO_LOAD>
+    ```
 
-Make sure that `--require` points to the main entry point for your app. If integrating with a Rails app, just point it to the directory of your application and it will handle loading the correct files on its own.
+    Make sure that `--require` points to the main entry point for the application. If integrating with a Rails app,
+    point it to the application directory and Chore will handle loading the correct files on its own.
 
-Other options include:
+1. When using SQS, ensure that AWS credentials exist in the environment (e.g. but not limited to `AWS_ACCESS_KEY_ID` &
+`AWS_SECRET_ACCESS_KEY` environment variables) and an AWS region is set (e.g. `AWS_REGION` environment variable) so that
+Chore can authenticate with AWS.
 
-    --concurrency 16 # number of concurrent worker processes, if using forked worker strategy
-    --worker-strategy Chore::Strategy::ForkedWorkerStrategy # which worker strategy class to use
-    --consumer Chore::Queues::SQS::Consumer # which consumer class to use Options are SQS::Consumer and Filesystem::Consumer. Filesystem is recommended for local and testing purposes only.
-    --consumer-strategy Chore::Queues::Strategies::Consumer::ThreadedConsumerStrategy # which consuming strategy to use. Options are SingleConsumerStrategy and ThreadedConsumerStrategy. Threaded is recommended for better tuning your consuming profile
-    --consumer-sleep-interval 1.0 # The amount of time in seconds to sleep when a consumer doesn't receive any messages. Sub-second values are accepted. The default varies by consumer implementation. This is a weak form of backoff for when there is no work to do.
-    --threads-per-queue 4 # number of threads per queue for consuming from a given queue.
-    --dedupe-servers # if using SQS or similiar queue with at-least once delivery and your memcache is running on something other than localhost
-    --batch-size 50 # how many messages are batched together before handing them to a worker
-    --batch-timeout 20 # maximum number of seconds to wait until handing a message over to a worker
-    --queue_prefix prefixy # A prefix to prepend to queue names, mainly for development and qa testing purposes
-    --max-attempts 100 # The maximum number of times a job can be attempted
-    --dupe-on-cache-failure # Determines the deduping behavior when a cache connection error occurs. When set to `false`, the message is assumed not to be a duplicate. Defaults to `false`.
-    --queue-polling-size 10 # If your particular queueing system supports responding with messages in batches of a certain size, you can control that with this flag. SQS has a built in upper-limit of 10, but other systems will vary.
+    By default, Chore will run over all queues it detects among the required files. If different behavior is desired,
+    use one of the following flags:
 
-If you're using SQS, you'll want to add AWS keys so that Chore can authenticate with AWS.
-
-    --aws-access-key=<AWS KEY>
-    --aws-secret-key=<AWS SECRET>
-
-By default, Chore will run over all queues it detects among the required files. If you wish to change this behavior, you can use:
-
+    ```
+    # Note that only one of these options may be used, not both. Chore will quit
+    # if both options are specified.
     --queues QUEUE1,QUEUE2... # a list of queues to process
     --except-queues QUEUE1,QUEUE2... # a list of queues _not_ to process
+    ```
 
-Note that you can use one or the other but not both. Chore will quit and make fun of you if both options are specified.
+1. Chore has many more options, which can be viewed by executing `bundle exec chore --help`
 
-### Tips for configuring Chore
+### Tips For Configuring Chore
 
-For Rails, it can be necessary to add the directory you store your jobs in to the eager loading path,
-found in application.rb. You likely need a similar approach for most apps using jobs, unless you places
-them into a directory that is already eager loaded by default. One example of this might be:
+For Rails, it can be necessary to add the jobs directory to the eager loading path, found in `application.rb`. A similar
+approach for most apps using jobs is likely needed, unless the jobs are placed into a directory that is already eager
+loaded by the application. One example of this might be:
 
 ```ruby
 config.eager_load_paths += File.join(config.root, "app", "jobs")
 ```
 
-However, due to the way eager_load_paths works in Rails, this may only solve the issue
-in your production environment. You might also find it useful for other environments
-to throw soemthing like this in an config/initializers/chore.rb file, although you
-can choose to load the job files in any way you like:
+However, due to the way `eager_load_paths` works in Rails, this may only solve the issue in the production environment.
+It can also be useful useful for other environments to have something like this in an `config/initializers/chore.rb`
+file, although the job files can be loaded in just about any way.
 
 ```ruby
-Dir["#{Rails.root}/app/jobs/**/*"].each do |file|
-  require file unless File.directory?(file)
-end unless Rails.env.production?
+if !Rails.env.production?
+  Dir["#{Rails.root}/app/jobs/**/*"].each do |file|
+    require file unless File.directory?(file)
+  end
+end
 ```
 
-### Producing and Consuming Jobs
+### Producing & Consuming Jobs
 
-When it comes to configuring Chore, you have 2 main use cases - as a producer of messages, or as a consumer of messages (the consumer is also able to produce messages if need be, but is running as it's own isolated instance of your application).
+When it comes to configuring Chore, there are 2 main use configurations - as a producer of messages, or as a consumer of
+messages. The consuming context may also messages if necessary, as it is running as its own isolated instance of the
+application.
 
-For producers, you must do all of your Chore configuration in an intializer.
+For producers, all of the Chore configuration must be in an initializer.
 
-For consumers, you need to either use a Chorefile or Chorefile + an initializer.
+For consumers, a Chorefile must be used. A Chorefile _plus_ an initializer is also a good pattern.
 
-Because you are likely to use the same app as the basis for both producing and consuming messages, you'll already have a considerable amount of configuration in your Producer - it makes sense to use Chorefile to simply provide the `require` option, and stick to the initializer for the rest of the configuration to keep things DRY.
-
-However, like many aspects of Chore, it is ultimately up to the developer to decide which use case fits their needs best. Chore is happy to let you configure it in almost any way you want.
-
-An example of how to configure chore via an initializer:
+Here's example of how to configure chore via an initializer:
 
 ```ruby
 Chore.configure do |c|
@@ -99,27 +97,42 @@ Chore.configure do |c|
 end
 ```
 
+Because it is like that the same application serves as the basis for both producing and consuming messages, and there
+will already be a considerable amount of configuration in the Producer, it makes sense to use Chorefile to simply
+provide the `require` option and stick to the initializer for the rest of the configuration to keep things DRY.
+
+However, like many aspects of Chore, it is ultimately up to the developer to decide which use case fits their needs
+best. Chore is happy to be configured in almost any way a developer desires.
+
 ## Integration
 
-Add an appropriate line to your `Procfile`:
+This section assumes `foreman` is being used to execute (or export the run commands of) the application, but it is not
+ strictly necessary.
 
+1. Add an appropriate line to the `Procfile`:
+
+    ```
     jobs: bundle exec chore -c config/chore.config
+    ```
 
-If your queues do not exist, you must create them before you run the application:
+1. If the queues do not exist, they must be created before the application can produce/consume Chore jobs:
 
-```ruby
-require 'aws-sdk'
-sqs = AWS::SQS.new
-sqs.queues.create("test_queue")
-```
+    ```ruby
+    require 'aws-sdk-sqs'
+    sqs = Aws::SQS::Client.new
+    sqs.create_queue(queue_name: "test_queue")
+    ```
 
-Finally, start foreman as usual
+1. Finally, start the application as usual
 
+    ```
     bundle exec foreman start
+    ```
 
-## Chore::Job
+## `Chore::Job`
 
-A Chore::Job is any class that includes `Chore::Job` and implements `perform(*args)` Here is an example job class:
+A `Chore::Job` is any class with `include Chore::Job` and implements a `perform(*args)` instance method. Here is an
+example job class:
 
 ```ruby
 class TestJob
@@ -129,30 +142,36 @@ class TestJob
   def perform(args={})
     Chore.logger.debug "My first async job"
   end
-
 end
 ```
 
-This job declares that the name of the queue it uses is `test_queue`, set in the queue_options method.
+This job declares that the name of the queue it uses is `test_queue`, set in the `queue_options` method.
 
-### Chore::Job and perform signatures
+### `Chore::Job` & `perform` Signatures
 
-The perform method signature can have explicit argument names, but in practice this makes changing the signature more difficult later on. Once a Job is in production and is being used at a constant rate, it becomes problematic to begin mixing versions of jobs which have non-matching signatures.
+The perform method signature can have explicit argument names, but in practice this makes changing the signature more
+difficult later on. Once a `Chore::Job` is in production and being used at a constant rate, it becomes problematic to
+begin mixing versions of the job with non-matching signatures.
 
-While this is able to be overcome with a number of techniques, such as versioning your jobs/queues, it increases the complexity of making changes.
+While this is able to be overcome with a number of techniques, such as versioning jobs/queues, it increases the
+complexity of making changes.
 
-The simplest way to structure job signatures is to treat the arguments as a hash. This will allow you to maintain forwards and backwards compatibility between signature changes with the same job class.
+The simplest way to structure job signatures is to treat the arguments as a hash. This enables maintaining forwards and
+backwards compatibility between signature changes with the same job class.
 
-However, Chore is ultimately agnostic to your particular needs in this regard, and will let you use explicit arguments in your signatures as easily as you can use a simple hash - the choice is left to you, the developer.
+However, Chore is ultimately agnostic in this regard and will allow explicit arguments in signatures as easily as using
+a simple hash; the choice is left to the developer.
 
-### Chore::Job and publishing Jobs
+### `Chore::Job` & Publishing Jobs
 
-Now that you've got a test job, if you wanted to publish to that job it's as simple as:
+Now that there's a test job, publishing an instance of the job is as simple as:
+
 ```ruby
 TestJob.perform_async({"message"=>"YES, DO THAT THING."})
 ```
 
-It's advisable to specify the Publisher chore uses to send messages globally, so that you can change it easily for local and test environments. To do this, you can add a configuration block to an initializer like so:
+It's advisable to specify the Publisher Chore uses to send messages globally, so that it can easily be modified based on
+the environment. To do this, add a configuration block to an initializer:
 
 ```ruby
 Chore.configure do |c|
@@ -160,14 +179,15 @@ Chore.configure do |c|
 end
 ```
 
-It is worth noting that any option that can be set via config file or command-line args can also be set in a configure block.
+It is worth noting that any option that can be set via config file or command-line args can also be set in a configure
+block.
 
 If a global publisher is set, it can be overridden on a per-job basis by specifying the publisher in `queue_options`.
 
 ## Retry Backoff Strategy
 
-Chore has basic support for delaying retries  of a failed job using a step function. Currently the only queue that
-supports this functionality is SQS, all others will simply ignore the delay setting.
+Chore has basic support for delaying retries of a failed job using a step function. Currently the only queue that
+supports this functionality is SQS; all others will simply ignore the delay setting.
 
 ### Setup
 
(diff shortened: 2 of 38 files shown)


lib/chore/consumer.rb
```diff
@@ -11,6 +11,13 @@ class Consumer
 
     attr_accessor :queue_name
 
+    # Raise this exception if your message has been processed but you can't delete it from the queue.
+    class CouldNotComplete < Exception; end
+
+    # @param [String] queue_name Name of queue to be consumed from
+    # @param [Hash] opts
+    #
+    # @return [Chore::Consumer] if so
```
alanbrent, line 20: Unnecessary

alanbrent: Also add yard-style documentation to the rest of the interface methods

lib/chore/publisher.rb
```diff
@@ -19,6 +19,12 @@ def self.publish(queue_name,job)
     def publish(queue_name,job)
       raise NotImplementedError
     end
+
```
alanbrent, line 22: Add YARD-style docs to methods in this interface.

lib/chore/queues/sqs.rb
```diff
@@ -1,6 +1,10 @@
 module Chore
   module Queues
     module SQS
+      def self.sqs_client
```
alanbrent, line 4: Add YARD-style documentation to this file

lib/chore/consumer.rb
```diff
@@ -63,5 +70,13 @@ def dupe_detector
       @dupes ||= DuplicateDetector.new({:servers => Chore.config.dedupe_servers,
                                         :dupe_on_cache_failure => false})
     end
+
+    private
+
+    def handle_messages(&block)
+      raise NotImplementedError
+      # Should do a block.call(id, receipt_id, queue_name, queue_timeout, job_json, previous_attempts)
```
alanbrent, line 78: Make this language better

lib/chore/queues/sqs/consumer.rb
```diff
@@ -43,21 +48,39 @@ def consume(&handler)
...
+          raise Chore::Consumer::CouldNotComplete, 'Required param "receipt_handle" missing!' unless opts[:receipt_handle]
+          queue.delete_messages(entries: [{ id: id, receipt_handle: opts[:receipt_handle] }])
         end
 
+        # Delays retry of a job by #{backoff_calc} seconds.
+        #
+        # @param [UnitOfWork] item Item to be delayed
+        # @param [Proc?] backoff_calc Code that determines the backoff.
+        #
+        # @return [Numeric]
```
alanbrent, line 76: Integer

lib/chore/queues/sqs/consumer.rb
```diff
@@ -17,24 +14,32 @@ class Consumer < Chore::Consumer
...
 
         # Sets a flag that instructs the publisher to reset the connection the next time it's used
+        #
+        # @return [Array<Seahorse::Client::NetHttp::ConnectionPool>]
         def self.reset_connection!
           @@reset_at = Time.now
+          Aws.empty_connection_pools!
         end
 
         # Begins requesting messages from SQS, which will invoke the +&handler+ over each message
```
alanbrent, line 32: Update description language

lib/chore/queues/sqs/consumer.rb
```diff
@@ -43,21 +48,39 @@ def consume(&handler)
...
           Chore.logger.debug "Completing (deleting): #{id}"
-          queue.batch_delete([id])
+          raise Chore::Consumer::CouldNotComplete, 'Required param "receipt_handle" missing!' unless opts[:receipt_handle]
+          queue.delete_messages(entries: [{ id: id, receipt_handle: opts[:receipt_handle] }])
         end
 
+        # Delays retry of a job by #{backoff_calc} seconds.
+        #
+        # @param [UnitOfWork] item Item to be delayed
+        # @param [Proc?] backoff_calc Code that determines the backoff.
```
alanbrent, line 74: Proc

lib/chore/queues/sqs/consumer.rb
```diff
@@ -66,46 +89,62 @@ def delay(item, backoff_calc)
...
         def handle_messages(&block)
-          msg = queue.receive_messages(:limit => sqs_polling_amount, :attributes => [:receive_count])
+          msg = queue.receive_messages(:max_number_of_messages => sqs_polling_amount, :attribute_names => ['ApproximateReceiveCount'])
           messages = *msg
+
           messages.each do |message|
-            unless duplicate_message?(message.id, message.queue.url, queue_timeout)
-              block.call(message.handle, queue_name, queue_timeout, message.body, message.receive_count - 1)
+            unless duplicate_message?(message.message_id, message.queue_url, queue_timeout)
+              # This provides data necessary for the worker to populate the UnitOfWork struct
```
alanbrent, line 102: Improve language and move to interface

lib/chore/strategies/consumer/single_consumer_strategy.rb
```diff
@@ -12,14 +12,18 @@ def initialize(fetcher, opts={})
       # Begins fetching from the configured queue by way of the configured Consumer. This can only be used if you have a
       # single queue which can be kept up with at a relatively low volume. If you have more than a single queue configured,
       # it will raise an exception.
+      #
+      # @param [Proc] &handler Message handler
+      #
+      # @return [Numeric] if so
```
alanbrent, line 18: What

lib/chore/unit_of_work.rb
```diff
@@ -19,7 +20,7 @@ def initialize(*) #:nodoc:
 
     # The current attempt number for the worker processing this message.
     def current_attempt
-      previous_attempts + 1
+      previous_attempts.to_i + 1
```
alanbrent, line 23: Call out for review (later)

alanbrent, line 23: I came across this and it seemed like a subtle bug. Can anyone think of a reason _not_ to do this?

lib/chore/queues/sqs/publisher.rb
```diff
@@ -3,51 +3,64 @@
...
+        # @return [struct Aws::SQS::Types::SendMessageResult]
         def publish(queue_name,job)
-          queue = self.queue(queue_name)
-          queue.send_message(encode_job(job))
+          queue = queue(queue_name)
+          queue.send_message(message_body: encode_job(job))
         end
 
         # Sets a flag that instructs the publisher to reset the connection the next time it's used
+        # TODO: We reset on every publish. Is this desirable? Seems like a performance problem.
```
alanbrent, line 29: Calling this out. I don't think we should fix it in this PR, but I'd like to know if we should leave the TODO in.

spec/chore/queues/sqs/consumer_spec.rb
```diff
@@ -1,145 +1,195 @@
...
+          expect(consumer.send(:queue)).to_not be_nil
+          consume
+        end
+
+        # TODO: This seems like it's no longer necessary. If it is, fix and reenable it, otherwise delete
+        xit 'should look up the queue based on the queue url' do
+          expect(sqs).to receive(:get_queue_url).with(:queue_name=>queue_name).and_return(queue_url)
+          expect(queue).to receive(:[]).with(queue_url).and_return(queue)
+          consume
+        end
```
alanbrent, line 73: Looking for opinions

spec/chore/queues/sqs/publisher_spec.rb
```diff
@@ -1,73 +1,63 @@
...
-      it 'should reconfigure sqs' do
-        Chore::Queues::SQS::Publisher.reset_connection!
-        AWS::SQS.should_receive(:new)
-        publisher.queue(queue_name)
+
+      # TODO: AThe reset_connection! method is setup so every call resets it, so this spec shouldn't even exist right?
+      # Also, the test itself seems like basic identity (i.e. not even a real test)
+      xit 'should not reset the connection between calls' do
+        sqs = publisher.send(:queue, queue_name)
+        expect(sqs).to be publisher.send(:queue, queue_name)
```
alanbrent, line 60: Looking for opinions

alanbrent:
Submitted to ops team for a first pass review. Will expand out to more people when this has passed through that filter.

README.md
```diff
@@ -1,92 +1,80 @@
...
 ```
 
-### Producing and Consuming Jobs
+### Producing & Consuming Jobs
 
-When it comes to configuring Chore, you have 2 main use cases - as a producer of messages, or as a consumer of messages (the consumer is also able to produce messages if need be, but is running as it's own isolated instance of your application).
+When it comes to configuring Chore, you have 2 main use cases - as a producer of messages, or as a consumer of messages.
 
-For producers, you must do all of your Chore configuration in an intializer.
+**IS THIS TRUE?** => (the consumer is also able to produce messages if need be, but is running as its own isolated instance of your application).
```
alanbrent, line 71: This doesn't seem correct to me, should I remove it?

obrie:
@alanbrent Could you provide a more comprehensive overview of what changes were made and why they were made?  It would be helpful to know what were QOL changes, what were bug fixes, what changes were made to explicitly support the AWS upgrade, etc. etc.

It *seems* like there's a lot of stuff in this PR that's not specifically for the AWS upgrade, but I'm not certain.

As a side note, is it possible to split out the changes into logical commits if they are unrelated to each other?

alanbrent:
@obrie I've updated the top comment as well as re-carved the commit history. Hopefully this helps, and definitely let me know what else you'd like to see in the top comment!

README.md
geoffreyclark: I assume this is just saying one _could_ call the `perform` of another job from within the consumer which seems plausible?

lib/chore/queues/sqs/publisher.rb
```diff
@@ -39,11 +40,7 @@ def self.reset_connection!
         #
         # @return [Aws::SQS::Client]
         def sqs
-          @sqs ||= AWS::SQS.new(
-            :access_key_id => Chore.config.aws_access_key,
-            :secret_access_key => Chore.config.aws_secret_key,
-            :logger => Chore.logger,
```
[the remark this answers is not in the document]
alanbrent: Good catch @geoffreyclark, I have added it back and also wired up the log level to match the Chore config log level.

README.md
alanbrent: A _worker_ can do that, but I don't see any logic in the consumer that would actually publish a message. @obrie do you happen to be able to figure out what was meant by this language?

lib/chore/unit_of_work.rb
obrie: Under what circumstances is this not an integer?  As far as I can tell, both the filesystem and sqs consumer pass an integer through here.  If the specs are passing a string, then I'd say the specs are wrong.

lib/chore/queues/sqs/publisher.rb
obrie: Can you point to where this is reset on every *publish* (just so we're using the same vocab, "publish" meaning writing a message to the sqs or filesystem queue).  The only place I see us resetting the connection is after forking, which is required.

lib/chore/unit_of_work.rb
alanbrent: In reality we never pass in anything but an integer as far as I can tell. I noticed this because by _looking_ at `UnitOfWork` it appears that `:previous_attempts` could at the very least be `nil`. But yeah, specs should cover that.

obrie: I think in that case we should remove this change as I see that as both a Ruby language design problem and a failed contract for the consumer.  I think better documentation could make it clearer what is expected to be nil vs. not nil.

alanbrent: _nod_ I will back out the change as part of addressing forthcoming feedback :)

lib/chore/queues/sqs/publisher.rb
alanbrent: We are setting `@@reset_next = true` every time `reset_connection!` is called

https://github.com/Tapjoy/chore/blob/04cd7a8745de51403f20ca9fa8b6f209b3bdce7b/lib/chore/queues/sqs/publisher.rb#L32-L35

Which means we will reset the connection every time `queue` is called

https://github.com/Tapjoy/chore/blob/04cd7a8745de51403f20ca9fa8b6f209b3bdce7b/lib/chore/queues/sqs/publisher.rb#L55-L60

And we call `queue` on every publish

https://github.com/Tapjoy/chore/blob/04cd7a8745de51403f20ca9fa8b6f209b3bdce7b/lib/chore/queues/sqs/publisher.rb#L23-L26

alanbrent: Aaaand now I see the `@@reset_next = false` ... so nevermind! My brain looked right past that.

lib/chore/queues/sqs/publisher.rb
```diff
@@ -3,51 +3,64 @@
...
+          queue.send_message(message_body: encode_job(job))
         end
 
         # Sets a flag that instructs the publisher to reset the connection the next time it's used
+        # TODO: We reset on every publish. Is this desirable? Seems like a performance problem.
+        #
+        # @return [Array<Seahorse::Client::NetHttp::ConnectionPool>]
         def self.reset_connection!
           @@reset_next = true
+          Aws.empty_connection_pools!
```
obrie, line 34: Just noting that this is also impacted by the same comment on where the logic should live

lib/chore/strategies/consumer/single_consumer_strategy.rb
```diff
@@ -10,16 +10,18 @@ def initialize(fetcher, opts={})
       end
 
       # Begins fetching from the configured queue by way of the configured Consumer. This can only be used if you have a
-      # single queue which can be kept up with at a relatively low volume. If you have more than a single queue configured,
-      # it will raise an exception.
+      # single queue which can be kept up with at a relatively low volume. If you have more than a single queue
+      # configured, it will raise an exception.
+      #
+      # @return [TrueClass, FalseClass]
```
obrie, line 16: No return doc needed here (I'll stop repeating myself -- just need to apply this across the doc changes you've made)

lib/chore/strategies/consumer/throttled_consumer_strategy.rb
```diff
@@ -100,22 +100,21 @@ def start_consumer_thread(consumer)
...
-          work = UnitOfWork.new(id, queue, timeout, body,
-                                previous_attempts)
+          work = UnitOfWork.new(message_id, message_receipt_handle, queue, timeout, body, previous_attempts)
           Chore.run_hooks_for(:consumed_from_source, work)
           @queue.push(work) if running?
           Chore.run_hooks_for(:added_to_queue, work)
         end
       end
-    end # ThrottledConsumerStrategyyeah 
+    end # ThrottledConsumerStrategy
```
obrie, line 118: :laughing: I do like the idea of a Lil' Jon language where the classes all end in `yeah`

https://www.youtube.com/watch?v=Wx3-cfE-6is

lib/chore/strategies/worker/helpers/ipc.rb
```diff
@@ -75,7 +75,7 @@ def ipc_help
 
       private
 
-      # TODO: do we need this as a optional param
+      # TODO: Decide if we should make this customizable via an optional param
```
obrie, line 78: I think the answer is probably "no" since we haven't needed to do it in years :laughing:

CHANGELOG.md
```diff
@@ -1,15 +1,22 @@
 # Change Log
 
-## [master](https://github.com/Tapjoy/chore/tree/master)
+## [v4.0.0](https://github.com/Tapjoy/chore/tree/v4.0.0)
```
obrie, line 3: Can you create a new `[master]` section as part of this?

docs/Delayed Jobs.md
```diff
@@ -1,28 +1,17 @@
...
-counting down from the moment the message is *first* sent, and does not reset.
+Something else to consider is the `RetentionPeriod` of a queue. If the `RetentionPeriod` is 1 hour, and an initial message continues to retry through that hour, it will suddenly disappear! That is to say: the `RetentionPeriod` starts counting down from the moment the message is *first* sent, and does not reset.
 
-And finally, if the queue has a `DeliveryDelay` configured any delay applied by the backoff function will be *in
-addition* to the `DeliveryDelay` when the message is removed from its "in flight" status.
+Finally, if the queue has a `DeliveryDelay` configured any delay applied by the backoff function will be *in addition* to the `DeliveryDelay` when the message is removed from its "in flight" status.
 
-All of this should be taken into consideration when using this feature. In general, any queue requiring this will likely
-need to run as its own Facet and not share workers with other queues.
+All of this should be taken into consideration when using this feature. In general, any queue requiring this will likely need to run as its own Facet and not share workers with other queues.
```
obrie, line 11: Could you go back to limiting these lines to 80 / 100-ish characters?  It becomes pretty difficult to read these files when I have to horizontally scroll each line.  I'm sure you can set up the editor to do this for you, but I'd prefer to not assume what editor settings people have, particularly for open-source projects.  Wrapped lines in Markdown is fairly common because Markdown will render it the same either way.

Examples:
* https://raw.githubusercontent.com/rails/rails/master/README.md
* https://raw.githubusercontent.com/lsegal/yard/main/README.md

lib/chore/queues/sqs/consumer.rb
```diff
@@ -43,21 +48,39 @@ def consume(&handler)
...
+        # Deletes the given message from the SQS queue
+        #
+        # @param [String] message_id Unique ID of the SQS message
+        # @param [Hash] opts Options, must include :receipt_handle (unique per consume request) of the SQS message
+        #
+        # @return [struct<Aws::SQS::Types::DeleteMessageBatchResult>]
+        def complete(message_id, opts = {})
           Chore.logger.debug "Completing (deleting): #{id}"
-          queue.batch_delete([id])
+          raise Chore::Consumer::CouldNotComplete, 'Required param "receipt_handle" missing!' unless opts[:receipt_handle]
```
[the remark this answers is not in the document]
obrie: On second thought, maybe we should just make it a standard second parameter.  That's probably simpler and doesn't give the impression that you can just easily pass through any generic options through here.  It also reflects the fact that this data is always passed in.

spec/chore/queues/spec_helper.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+module Chore
+  RSpec.shared_context 'fake objects' do
+    let(:job) { {'class' => 'TestJob', 'args'=>[1,2,'3']}}
+    let(:queue_name) { 'test_queue' }
+    let(:queue_url) { "http://amazon.sqs.url/queues/#{queue_name}" }
+
+    let(:queue) do double(Aws::SQS::Queue,
```
obrie, line 9: Missing a newline between `do` and `double`

spec/chore/queues/sqs_spec.rb
```diff
@@ -1,35 +1,28 @@
-require 'spec_helper'
+require_relative 'spec_helper'
```
obrie, line 1: Curious what led to needing this change?

spec/chore/queues/spec_helper.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+module Chore
+  RSpec.shared_context 'fake objects' do
+    let(:job) { {'class' => 'TestJob', 'args'=>[1,2,'3']}}
```
obrie, line 5: This seems like an outlier that doesn't need to be in the shared context.  I can see an argument for the more complex s3 class creation being shared because you're effectively creating a fake_sqs, though.

spec/chore/queues/sqs/consumer_spec.rb
```diff
@@ -1,145 +1,194 @@
...
+    let(:receive_message_result) { Aws::SQS::Message::Collection.new([message], size: 1) }
+
+    let(:message) do
+       Aws::SQS::Message.new(
+        message_id: 'message id',
+        receipt_handle: "receipt_handle",
+        body: job.to_json,
+        data: job,
+        queue: queue,
+        queue_url: sqs.get_queue_url.queue_url,
```
obrie, line 22: This seems like we're taking advantage of the fact that the `get_queue_url` method doesn't actually require any arguments which I think isn't actually the case and is a bit confusing.  I think it'd be clearer to reference the `queue_url` variable here.

spec/chore/queues/sqs/consumer_spec.rb
```diff
@@ -1,145 +1,194 @@
...
+      consumer.consume(&block)
     end
 
-    it 'should look up the queue based on the queue url' do
-      expect(queues).to receive(:[]).with(queue_url).and_return(queue)
-      consumer.consume
+    before do
+      allow(Aws::SQS::Client).to receive(:new).and_return(sqs)
+      allow(Aws::SQS::Queue).to receive(:new).and_return(queue)
+      allow(sqs).to receive(:receive_message).and_return(receive_message_result)
```
obrie, line 36: Why are we stubbing this on `sqs` instead of just directly on the queue?  Do we actually Chore to call `Aws::SQS::Client#receive_message`?

spec/chore/queues/sqs/consumer_spec.rb
```diff
@@ -1,145 +1,194 @@
...
-      consumer.consume
+    # Since a message handler is required (but not validated), this convenience method lets us
+    # effectively stub the block.
+    def consume(&block)
+      block = Proc.new{} unless block_given?
+      consumer.consume(&block)
     end
 
-    it 'should look up the queue based on the queue url' do
-      expect(queues).to receive(:[]).with(queue_url).and_return(queue)
```
obrie, line 55: Just for context, I believe the reason the `and_return` was being done here is because it would (at least on older versions of rspec), by default, return `nil` if a previous `allow(...)...and_return(...)` wasn't defined for the same object/method.  This would cause problems with the caller.

lib/chore/queues/sqs/consumer.rb
```diff
@@ -66,46 +89,59 @@ def delay(item, backoff_calc)
...
+            self.class.reset_connection!
             @sqs = nil
             @sqs_last_connected = Time.now
             @queue = nil
           end
-          @queue_url ||= sqs.queues.url_for(@queue_name)
-          @queue ||= sqs.queues[@queue_url]
+
+          @queue_url ||= sqs.get_queue_url(queue_name: @queue_name).queue_url
+          @queue ||= Aws::SQS::Queue.new(url: @queue_url)
```
obrie, line 125: Looks like we should be passing in `client: sqs` here: https://github.com/aws/aws-sdk-ruby/blob/877d729fa4176891e68c77143b80a3ae6612b9e7/gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue.rb#L16-L26

Example of how the Resource class builds queues: https://github.com/aws/aws-sdk-ruby/blob/877d729fa4176891e68c77143b80a3ae6612b9e7/gems/aws-sdk-sqs/lib/aws-sdk-sqs/resource.rb#L260-L269

spec/chore/queues/sqs/consumer_spec.rb
```diff
@@ -1,145 +1,194 @@
...
-    end
 
-    it "should check the uniqueness of the message" do
-      allow_any_instance_of(Chore::DuplicateDetector).to receive(:found_duplicate?).with(message_data).and_return(false)
-      consumer.consume
-    end
+      context "should receive a message from the queue" do
+        before do
+          allow(consumer).to receive(:queue).and_return(queue)
+        end
```
obrie, line 79: Is this required if you've previously stubbed it with `allow(Aws::SQS::Queue).to receive(:new).and_return(queue)`?

spec/chore/queues/sqs/publisher_spec.rb
```diff
@@ -1,73 +1,63 @@
...
       publisher.publish(queue_name,job)
     end
 
-    it 'should create send an encoded message to the specified queue' do
-      queue.should_receive(:send_message).with(job.to_json)
-      publisher.publish(queue_name,job)
+    it 'should create a new SQS client before every publish' do
+      expect(Aws::SQS::Client).to receive(:new).twice
+      publisher.send(:queue, queue_name)
+      publisher.send(:queue, queue_name)
```
obrie, line 23: Can be `2.times { publisher.send(:queue, queue_name) }`

spec/chore/queues/sqs/publisher_spec.rb
obrie: I think this is still a spec we'll want

README.md
obrie: I think it's just talking about the consuming application, not the consumer queue implementation.

lib/chore/queues/sqs/consumer.rb
```diff
@@ -66,46 +89,59 @@ def delay(item, backoff_calc)
...
+        # If <tt>reset_connection!</tt> has been called, this will result in the connection being re-initialized,
+        # as well as clear any cached results from prior calls
+        #
+        # @return [Aws::SQS::Queue]
         def queue
           if !@sqs_last_connected || (@@reset_at && @@reset_at >= @sqs_last_connected)
-            AWS::Core::Http::ConnectionPool.pools.each do |p|
-              p.empty!
-            end
+            self.class.reset_connection!
```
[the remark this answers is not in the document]
alanbrent: _nod_ I plan to do the first part (revert to previous flow), but the call defining that loop is no longer a thing. AFAICT the `Aws.reset_connection_pools!` does the same conceptual work.

```ruby
[1] pry(main)> show-source Aws.empty_connection_pools!

From: /Users/brent.stephens/Documents/development/src/github.com/tapjoy/chore/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-3.105.0/lib/aws-sdk-core.rb:153:
Owner: #<Class:Aws>
Visibility: public
Signature: empty_connection_pools!()
Number of lines: 5

def empty_connection_pools!
  Seahorse::Client::NetHttp::ConnectionPool.pools.each do |pool|
    pool.empty!
  end
end
```

spec/chore/queues/sqs_spec.rb
obrie: I understand now what the intent here -- discussed in other threads.

README.md
alanbrent: That makes _a lot_ more sense, thank you. I will update this language.

spec/chore/queues/sqs/consumer_spec.rb
```diff
@@ -1,145 +1,194 @@
...
-    end
-
-    it 'should not configure sqs multiple times' do
-      allow(consumer).to receive(:running?).and_return(true, true, false)
-
-      expect(AWS::SQS).to receive(:new).once.and_return(sqs)
-      consumer.consume
+require_relative '../spec_helper'
+
+module Chore
```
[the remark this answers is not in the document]
alanbrent: The goal was to share AWS API object doubles in an effort to reduce the cognitive overhead of reasoning about all the SQS-related specs together. I just cargo-culted `module Chore` because that's the context the specs were all running in. I will finish reading through your interrelated  comments and either make updates or ask more questions :)

lib/chore/queues/sqs/consumer.rb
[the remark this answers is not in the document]
alanbrent: Should I document anything around the signature being different than the one defined in the interface, or is that a common thing that won't surprise developers?

lib/chore/queues/sqs/consumer.rb
alanbrent: Good eye, thank you!

lib/chore/queues/sqs/consumer.rb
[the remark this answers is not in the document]
obrie: What do you think about just updating the interface?  Implementations can then decide whether they need/respect the argument.

alanbrent:
Alright I think we're ready for the next round. Thanks for all the feedback :)

lib/chore/queues/sqs/publisher.rb
```diff
@@ -3,51 +3,62 @@
...
         def publish(queue_name,job)
-          queue = self.queue(queue_name)
-          queue.send_message(encode_job(job))
+          queue = queue(queue_name)
+          queue.send_message(message_body: encode_job(job))
         end
 
         # Sets a flag that instructs the publisher to reset the connection the next time it's used
+        #
+        # @return [Array<Seahorse::Client::NetHttp::ConnectionPool>]
```
obrie, line 30: Out-of-date doc (No need to document a return value here)

spec/chore/queues/sqs/consumer_spec.rb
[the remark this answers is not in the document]
obrie: I'm not sure if I follow here.  What's the benefit in putting specs in a module?  As far as I'm aware, we don't do this in any of our codebases and I'm not sure I've ever seen that as a practice in rspec docs.  The `describe` blocks themselves create anonymous modules, so there isn't typically value in surrounding them with another module.

obrie (changes requested):
The question on rspec modules is the main outstanding comment I think on my end.

spec/chore/queues/sqs/consumer_spec.rb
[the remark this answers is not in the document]
alanbrent: The same-functionality-but-different-implementation fake objects between the publisher and consumer was confusing to me, and it seemed to make sense to unify the objects to reduce that confusion as well as prevent drift. As far as the enclosing module declaration in the specs themselves, that was pre-existing. Did you want me to remove them?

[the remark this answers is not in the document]
alanbrent: Update: it was pre-existing in the _publisher_, and I guess I updated the other spec to match.

obrie (dismissed):
I'll take one more pass through it, but this seems ready for a production test!

obrie (dismissed):
:partying_face: :champagne: To version infinity and beyond!

alanbrent:
Thanks @ehealy for catching a bug in the consumer whose fix is captured in the diff below. I have updated the PR with a fix as well as an additional spec to more fully test the `complete` method.

```diff
diff --git a/lib/chore/queues/sqs/consumer.rb b/lib/chore/queues/sqs/consumer.rb
index 7a76d49..4ca8ed5 100644
--- a/lib/chore/queues/sqs/consumer.rb
+++ b/lib/chore/queues/sqs/consumer.rb
@@ -58,8 +58,8 @@ module Chore
         # @param [String] message_id Unique ID of the SQS message
         # @param [Hash] receipt_handle Receipt handle (unique per consume request) of the SQS message
         def complete(message_id, receipt_handle)
-          Chore.logger.debug "Completing (deleting): #{id}"
-          queue.delete_messages(entries: [{ id: id, receipt_handle: receipt_handle }])
+          Chore.logger.debug "Completing (deleting): #{message_id}"
+          queue.delete_messages(entries: [{ id: message_id, receipt_handle: receipt_handle }])
         end
```

ehealy (approved):
Using https://github.com/Tapjoy/opsbot/pull/84 as a testbed, I have validated the following:

* SQS Producer
- [x] Initializes without issue
- [x] Successfully submits messages to SQS

* SQS Consumer utilizing the Threaded Consumer Strategy
- [x] Initializes without issue
- [x] Successfully polls SQS for batches of messages for the Chore jobs specified
- [x] Assigns messages to workers
- [x] Workers complete Chore jobs
- [x] Chore jobs are successfully removed form the appropriate SQS queues.

Any feedback/bugfixes were addressed via facewords and re-validated during testing cycle.

Outstanding work @alanbrent.

ehealy (approved):

PR #56, opened 2020-08-27
Consumer spec fixes

dabrady:
@alanbrent I did most of this in one large chunk (lots of trial-and-error on spec setup and such) but tried to split it into digestible chunks for illustrative porpoises.

Feel free to do with what you will, and we can chat tomorrow about anything if you want. All the specs pass on my machine:
image


alanbrent:
Thanks! I will merge in now, though fair warning I'm probably going to reconstruct the specs to match the patterns used in the SQS publisher specs.

PR #58, opened 2020-09-16
Remove appended underscore from CLI `--queue-prefix`

alanbrent:
The prefixing should be handled entirely by the consuming application, rather than Chore appending to it at the last second.

We started running into this when we (I) unified the queue prefixing logic between our application's `Chorefile` (consume) and the `Chore.config` (publish) we use in our Chore initializer. This change did not affect our production applications, because we don't use `--queue-prefix` at all in our production environments.

## Pre-Release Questions

- [x] Are we comfortable with 4.0.1 as a release version? Technically this could break people who have learned to assume the previous (imo slightly wrong) behavior
    - We'll use 4.1.0 instead.
- [x] Do we want to backport this to 3.x? If so, what's the best way to go about doing so?
    - No, we don't :)

---

Given the long history of this behavior, I am also open to _not_ doing this and instead heavily commenting the reasons why our prefixing logic differs between the Chorefile and the initializers in our applications.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,18 @@
 **Cleanups**
 - N/A
 
+## [v4.1.0](https://github.com/Tapjoy/chore/tree/v4.1.0)
+
+**BREAKING**
+
+This change requires changes if you are using `--queue-prefix` argument when running the `bin/chore.rb` binstub. See bug
+fix details for more info.
+
+**Fixed bugs**
+
+- `Chore::CLI` was appending an underscore to any passed in `--queue_prefix`. This should be handled by the consuming
+  application, and has been removed.
+
 ## [v4.0.0](https://github.com/Tapjoy/chore/tree/v4.0.0)
 
 **Features**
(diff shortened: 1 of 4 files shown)


normchan-tj (approved):
lgtm

obrie:
Let's bump this to 4.1.0.  No need to backport.

alanbrent:
@obrie Done!

CHANGELOG.md
```diff
@@ -11,6 +11,18 @@
 **Cleanups**
 - N/A
 
+## [v4.1.0](https://github.com/Tapjoy/chore/tree/v4.0.0)
```
obrie, line 14: 4.0.0 => 4.1.0

obrie (changes requested):

CHANGELOG.md
alanbrent: Doh, fixed!

obrie (approved):

PR #60, opened 2023-06-06
NPE-45 Add automation for running specs on M1 Macs

alanbrent:
Also constraining `dalli` gem version, as upstream has broken us.


eric-kansas:
Tested locally and `make dev-specs` passed 
NOTE: some tests do seem non-deterministic and might fail.... just rerun

eric-kansas (approved):

PR #65, opened 2024-03-25
NPE-283: Remove requirement that Dalli be used as the memcache client for duplicate detection

obrie:
Currently, chore requires that the `dalli` gem be used for duplicate detection.  In fact, it's actually requiring that you must use memcache *and* dalli.  This has created some friction in cases where we're trying to reduce the number of external dependencies we have in an an application.  For example, by forcing dalli, it means that we must have both the `memcached` gem and the `dalli` gem installed for every application that uses chore.

This changes the `chore` interface so that `dalli` is merely the default.  If a custom `dedupe_handler` is provided, then that handler function will be called with the same content as `dalli`.

This will allow us to set `config.dedupe_handler` on the chore config to use a different memcached client.

### DIFF
--- a/lib/chore/version.rb
+++ b/lib/chore/version.rb
@@ -1,7 +1,7 @@
 module Chore
   module Version #:nodoc:
     MAJOR = 4
-    MINOR = 2
+    MINOR = 3
     PATCH = 0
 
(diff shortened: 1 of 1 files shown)


ojameso (approved):
Makes sense.

PR #68, opened 2024-09-17
NPE-194: Modernize external gem dependencies

obrie:
https://jira.unity3d.com/browse/NPE-194

This makes several changes to how we're managing our gem dependencies for chore:

* Replace `json` with `multi_json` so that the application decides with json implementation to use
* Remove `thread` dependency as we're not relying on any of its functionality
* Relax version pinning on dependencies
* Remove unused development dependencies
* Update specs to rspec 3 syntax

This should make it easier for applications to keep their gem versions up to date over time.


ojameso (approved):
👍

PR #72, opened 2025-07-15
[NPE] Add GCP Pubsub Support

ojameso:
This adds GCP PubSub support. 

Adds both `Chore::Queues::PubSub::Consumer` and `Chore::Queues::PubSub::Producer`. this can be configured numerous ways. Interesting notes regarding configuration have been added to the README. 

## Breaking change
1. Moves AWS-SQS to dev-dependency and not runtime since we can now use either AWS or GCP.
2. Drops support for Ruby 2.7

## Nonbreaking change
1. Add PubSub support
2. Queue creation tasks now separate into create_sqs/create_pubsub tasks.

## Testing
- [x] TIAB - Was able to create a test job in tapjoyserver to read/write form the Queue.


lib/chore/queues/pubsub.rb
```diff
@@ -0,0 +1,146 @@
...
+      def self.pubsub_client
+        require 'google/cloud/pubsub'
+        
+        # Verify compatible version
+        begin
+          gem_version = Gem.loaded_specs['google-cloud-pubsub']&.version
+          if gem_version && gem_version < Gem::Version.new('3.0.0')
+            raise "google-cloud-pubsub version #{gem_version} is not supported. Please use version >= 3.0"
+          end
+        rescue => e
```
captainkevron: If an incompatible version is detected, do we just want to log a warning? Or do we want to terminate with an error?

lib/chore/queues/pubsub.rb
```diff
@@ -0,0 +1,146 @@
...
+          if existing.size > 0
+            raise <<-ERROR.gsub(/^\s+/, '')
+            We found topics/subscriptions that already exist! Verify your queue names or prefix are setup correctly.
+
+            The following queue names were found:
+            #{existing.join("\n")}
+            ERROR
+          end
+        end
+        client = pubsub_client
```
captainkevron, line 42: Any reason to use this local var and not just use `pubsub_client` directly?

lib/chore/queues/pubsub.rb
```diff
@@ -0,0 +1,146 @@
...
+      end
+
+      # Helper method to delete all known topics and subscriptions based on the list as provided by your configured Chore::Jobs
+      # This is meant to be invoked from a rake task, and not directly.
+      #
+      # @return [Array<String>]
+      def self.delete_queues!
+        raise RuntimeError.new('You must have at least one Chore Job before attempting to delete Pub/Sub topics') unless Chore.prefixed_queue_names.length > 0
+
+        client = pubsub_client
```
captainkevron, line 81: Same question here.

captainkevron (commented):
Couple small questions. I am not very familiar with the `Chore` architecture, so I mostly used the `SQS` implementation as a reference to compare against. Nice work on the `README.md` updates.

lib/chore/queues/pubsub.rb
ojameso: Yea, the `pubsub_client` isn't memoized so whenever you call `pubsub_client`, it creates a fresh instance which makes it lose any cached values, etc. In the consumer/producer files, this is put into an instance variable, but it didn't seem necessary to do so in this context.

lib/chore/queues/pubsub.rb
ojameso: ~~It depends. Chore might not be using pubsub, but the application might.~~
I just realized this is when we're actually trying to instantiate a pubsub_client. Unsure. Perhaps at this point, we do want to just terminate loudly since if not, it will continue on and break somewhere else.

lib/chore/queues/sqs.rb
obrie, line 1: Since we're removing the aws sdk as a runtime dependency, we should add the necessary `require` statements to `def self.sqs_client`

chore.gemspec
```diff
@@ -36,10 +36,11 @@ Gem::Specification.new do |s|
   s.summary = "Job processing... for the future!"
 
   s.add_runtime_dependency(%q<multi_json>, [">= 0"])
-  s.add_runtime_dependency(%q<aws-sdk-sqs>, [">= 1"])
   s.add_runtime_dependency(%q<logger>, [">= 0"])
   s.add_runtime_dependency(%q<ostruct>, [">= 0"])
   s.add_development_dependency(%q<rspec>, [">= 3"])
+  s.add_development_dependency(%q<aws-sdk-sqs>, [">= 1"])
+  s.add_development_dependency(%q<google-cloud-pubsub>, [">= 3.0", "< 4.0"])
```
obrie: Can we remove the strict upper dependency?  If Google releases 4.0, we have to update this code, re-release chore, updating the dependency in TJS, etc.  Part of the goal of last year's work was to remove all upper dependencies in order to reduce the upgrade cycle.

lib/chore/tasks/queues.task
```diff
@@ -1,18 +1,46 @@
...
+  task :create, :halt_on_existing do |t, args|
+    Rake::Task['chore:queues:create_sqs'].invoke(args[:halt_on_existing])
   end
 
-  desc "Remove all defined queues"
+  desc "Remove all defined SQS queues (legacy - use chore:queues:remove_sqs instead)"
   task :remove do
-    Chore::Queues::SQS.delete_queues!
+    Rake::Task['chore:queues:remove_sqs'].invoke
   end
```
obrie: I would suggest we get rid of these legacy tasks and force consumers to upgrade.  We're already doing a major version bump.

lib/chore/queues/pubsub.rb
obrie: Yeah let's terminate and not immediately rescue the exception.  If we're going to do version checks, I'd also suggest using `Google::Cloud::PubSub::VERSION` comparison rather than going through gem specs.

lib/chore/queues/pubsub.rb
```diff
@@ -0,0 +1,146 @@
...
+        begin
+          gem_version = Gem.loaded_specs['google-cloud-pubsub']&.version
+          if gem_version && gem_version < Gem::Version.new('3.0.0')
+            raise "google-cloud-pubsub version #{gem_version} is not supported. Please use version >= 3.0"
+          end
+        rescue => e
+          Chore.logger.warn "Could not verify google-cloud-pubsub version: #{e.message}" if defined?(Chore.logger)
+        end
+        
+        if self.project_id && self.credentials
```
obrie: These arguments shouldn't be needed.  We can default all of this at a global scope, which is similar to what we do for S3.  Reference: https://github.com/googleapis/google-cloud-ruby/blob/c8f06ac89584fbef5ee64817e6a1310bb2573584/google-cloud-pubsub/lib/google/cloud/pubsub.rb#L142-L154

This way, the application has complete control over all arguments that get inherited here.  Otherwise, it makes it look like we're limiting how you can configure PubSub (https://github.com/googleapis/google-cloud-ruby/blob/c8f06ac89584fbef5ee64817e6a1310bb2573584/google-cloud-pubsub/lib/google/cloud/pubsub.rb#L74-L86)

lib/chore/queues/pubsub.rb
```diff
@@ -0,0 +1,146 @@
...
+  module Queues
+    module PubSub
+      class << self
+        attr_accessor :project_id, :credentials
+        
+        # Configure PubSub settings with a block
+        def configure
+          yield self if block_given?
+        end
+      end
```
obrie: Assuming we agree with removing the arguments (instead relying on global pubsub defaults), we should remove all of this code as well

lib/chore/queues/pubsub.rb
```diff
@@ -0,0 +1,146 @@
...
+        Chore.prefixed_queue_names
+      end
+
+      # Collect a list of topics/subscriptions that already exist
+      #
+      # @return [Array<String>]
+      def self.existing_queues
+        Chore.prefixed_queue_names.select do |queue_name|
+          begin
+            client = pubsub_client
```
obrie: Can we move this to outside the loop so we don't have to re-create the client for every queue?

lib/chore/queues/pubsub.rb
```diff
@@ -0,0 +1,146 @@
...
+            #{existing.join("\n")}
+            ERROR
+          end
+        end
+        client = pubsub_client
+
+        Chore.prefixed_queue_names.each do |queue_name|
+          Chore.logger.info "Chore Creating Pub/Sub Topic and Subscription: #{queue_name}"
+          topic_path = client.topic_path(queue_name)
+          subscription_name = "#{queue_name}-sub"
```
obrie: I would suggest making `-sub` a constant in this file that all code can reference and maybe having a single method that can generate a subscription name from a topic.  Otherwise, any changes to this logic need to get repeated in multiple places.

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,172 @@
...
+        # Initialize the reset at on class load
+        @@reset_at = Time.now
+
+        Chore::CLI.register_option 'gcp_project_id', '--gcp-project-id PROJECT_ID', 'GCP Project ID for Pub/Sub' do |project_id|
+          Chore::Queues::PubSub.project_id = project_id
+        end
+        
+        Chore::CLI.register_option 'gcp_credentials', '--gcp-credentials PATH', 'Path to GCP service account credentials JSON file' do |credentials|
+          Chore::Queues::PubSub.credentials = credentials
+        end
```
obrie: Pretty sure these CLI options are completely unused in SQS.  I'm thinking we just rip it out (both for SQS and PubSub).

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,172 @@
...
+        # Resets the API client connection and provides @@reset_at so we know when the last time that was done
+        def self.reset_connection!
+          @@reset_at = Time.now
+        end
+
+        # @param [String] queue_name Name of GCP Pub/Sub topic
+        # @param [Hash] opts Options
+        def initialize(queue_name, opts={})
+          super(queue_name, opts)
+          @subscription_name = "#{queue_name}-sub"
```
obrie: I'm wondering whether the queue_name should actually be the subscription_name?  I guess that only makes sense from the consumer context and that the publisher only cares about the topic name.

In production, we're going to be defining the topics / subscriptions in KCL and not relying on the default `-sub` suffix.  With the current implementation, we end up with a delicate expectation that KCL definitions follow the same pattern expected by chore.

Could we determine the subscription name by finding the first subscription for a topic and using that instead of making an assumption about what the subscription name is?

Otherwise, we just need to make sure this is enforced in KCL and avoid repeating that assumption in as many places as possible.

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,172 @@
...
+
+        # Acknowledges the given message from the GCP Pub/Sub subscription
+        #
+        # @param [String] message_id Unique ID of the Pub/Sub message  
+        # @param [String] ack_id Acknowledgment ID of the Pub/Sub message
+        def complete(message_id, ack_id)
+          Chore.logger.debug "Completing (acknowledging): #{message_id}"
+          # Find the message by ack_id and acknowledge it
+          if msg = @current_messages&.find { |m| m.ack_id == ack_id }
+            msg.acknowledge!
```
obrie: :thinking: It seems like we shouldn't need to keep this `@current_message` object and instead use the subscriptions `acknowledge` API: https://github.com/googleapis/google-cloud-ruby/blob/c8f06ac89584fbef5ee64817e6a1310bb2573584/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb#L442-L448

This API accepts a list of ack ids, which we already have as an argument to this method.  So something like:

```
subscription.acknowledge(ack_id)
```

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,172 @@
...
+        # @param [UnitOfWork] item Item to be delayed
+        # @param [Block] backoff_calc Code that determines the backoff.
+        def delay(item, backoff_calc)
+          delay = backoff_calc.call(item)
+          Chore.logger.debug "Delaying #{item.id} by #{delay} seconds"
+
+          # Find the message and modify its ack deadline
+          if msg = @current_messages&.find { |m| m.ack_id == item.receipt_handle }
+            msg.modify_ack_deadline!(delay)
+          end
```
obrie: Similar note here -- we don't need to find the message in `@current_messages`.  We can use this API: https://github.com/googleapis/google-cloud-ruby/blob/c8f06ac89584fbef5ee64817e6a1310bb2573584/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb#L477-L482

```
subscriptions.modify_ack_deadline(delay, item.receipt_handle)
```

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,172 @@
...
+        def handle_messages(&block)
+          begin
+            verify_connection!
+          rescue => e
+            Chore.logger.error "There was a problem verifying the connection to the subscription: #{e.message}. Shutting down..."
+            raise Chore::TerribleMistake
+          end
+
+          messages = subscription.pull(max: max_messages)
+          @current_messages = messages
```
obrie: I think we'll end up getting rid of this, but I also believe there's a race condition here where the consumer is pulling new messages while older messages are still being processed.

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,172 @@
...
+          received_timestamp = Time.now
+
+          messages.each do |message|
+            unless duplicate_message?(message.message_id, @subscription_name, queue_timeout, received_timestamp)
+              # delivery_attempt is available but may be nil for older messages
+              attempt_count = (message.delivery_attempt || 1) - 1
```
obrie: Can you confirm that `delivery_attempt` is always present?  I noticed [this message](https://github.com/googleapis/google-cloud-ruby/blob/c8f06ac89584fbef5ee64817e6a1310bb2573584/google-cloud-pubsub/lib/google/cloud/pubsub/received_message.rb#L80-L81) in the pubsub library:

```
        # @return [Integer, nil] A delivery attempt value of `1` or greater, or `nil` if a dead letter policy is not set
        #   on the subscription.
```

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,172 @@
...
+          end
+
+          @subscription ||= pubsub.subscriber(@subscription_name)
+        end
+
+        # The ack deadline (in seconds) of the subscription
+        #
+        # @return [Integer]
+        def queue_timeout
+          @queue_timeout ||= subscription.deadline || 600 # Default to 10 minutes
```
obrie: Where is the default of 10 minutes coming from?  I'm seeing the following (https://cloud.google.com/pubsub/docs/subscription-properties#ack_deadline):

> The following are the values for the Acknowledgment deadline option:
> 
> Default value = 10 seconds
> Minimum value = 10 seconds
> Maximum value = 600 seconds

In fact, this is concerning.  SQS has a maximum visibility timeout of 12 hours -- this allows for long-running jobs that could take long to run.  For example, the visibility timeout for `v2_CacheOffersToS3` in SQS in 30 minutes.  We've hit the 15 min mark for that job in the last 3 month according to NewRelic.

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,172 @@
...
+        # @return [Google::Cloud::PubSub::Project]
+        def pubsub
+          @pubsub ||= Chore::Queues::PubSub.pubsub_client
+        end
+
+        # Maximum number of messages to retrieve on each request
+        #
+        # @return [Integer]
+        def max_messages
+          [pubsub_polling_amount, 1000].min  # Pub/Sub max is 1000
```
obrie: I don't think we should have such a large default.  The problem is that if you have multiple consumers spread across several servers, then a single consumer could take all 1,000 and have a large backlog in-memory while it waits for workers to pick up the jobs.  Meanwhile, the other servers could be sitting there doing nothing.

I'd rather not have any default here as it's entirely dependent on the consumer and we already have a default defined for `Chore.config.queue_polling_size`

lib/chore/queues/pubsub/consumer.rb
ojameso: Yikes, you're right. I have that on my list of comparisons that I just glossed over.
```
Visibility Timeout | 0 – 12 hours (VisibilityTimeout) | 10 sec – 10 min (ackDeadlineSeconds)
```
Hm, we'll have to think about this one or reach out to the Google team about it.

ojameso: We could technically push it out, by modifying the ack_deadline. But I'll need to find an example of this.
```
received_message.modify_ack_deadline!(600)
```

lib/chore/queues/pubsub/consumer.rb
ojameso: This makes sense. The `-sub` suffix is the default in google and so it was going off of this assumption, but discovery of the subscription via topic is a safer process. The only thing would be if there are multiple subscriptions for a given topic, this can get interesting. However, I'm supposing that we will not do such a thing for these queues. I'm going to mull on this for a bit.

lib/chore/queues/pubsub/consumer.rb
ojameso: Hm, not sure if I'm understanding correctly. Once a message gets pulled, google sets an ack_deadline and prevents it from being pulled again. Is there something we're doing that pulls messages again the in the same thread?

obrie: I'm referring to the `@current_messages` instance variable.  It gets overwritten as the consumer is in a loop pulling more messages from the queue whereas other threads would be coming back and attempting to read the `@current_messages` instance variable to find the ack_id, which has since changed.  I meant it probably wasn't relevant if we didn't need `@current_messages` anymore

lib/chore/queues/pubsub/consumer.rb
ojameso: Hm, not quite sure what you mean here. The code already makes the assumption that `delivery_attempt` may be `nil`? In this case, it will default to `1`. Am I missing something?

lib/chore/queues/pubsub/consumer.rb
ojameso: Ah, gotcha! Thanks for clarification.

lib/chore/queues/pubsub/consumer.rb
obrie: The documentation seems to indicate that it was *always* be `nil` if no dead letter policy is set (even when we're on the Nth delivery attempt).  If that's true, then this value will be incorrect -- it will always be `1` even if we're retrying for the 10th time.

lib/chore/queues/pubsub/consumer.rb
ojameso: I've attempted to address this by using a thread that loops through and pushes out the ack deadline. @obrie if you have thoughts, let me know. Separated that into its own commit so you can see it easily.

obrie: Hrm I think there's a misunderstanding with what's happening here and how to solve it.  Let's revert that change and talk through it over VC.

ojameso: I'm going to resolve this since we ar ehandling this outside of the scope of this PR.

lib/chore/queues/pubsub/consumer.rb
ojameso: After looking into, delivery_attempt seems to be only used to handle sqs backoff logic within chore. This has no application for pubsub and will simply default to the queue configuration where such exponential delay can be configured. Resolving as I don't see this impacting service in any meaningful way.

README.md
```diff
@@ -16,11 +16,24 @@ Chore can be integrated with any Ruby-based project by following these instructi
...
+    **Queue Provider Dependencies**: Add the appropriate gem for your queue provider:
 
+    For **Amazon SQS**:
+    ```ruby
+    gem 'aws-sdk-sqs'
+    ```
+
+    For **Google Cloud Pub/Sub** (requires Ruby 3.1+ and google-cloud-pubsub 3.0+):
+    ```ruby
+    gem 'google-cloud-pubsub', '>= 3.0', '< 4.0'
```
obrie: Is the `< 4.0` required?

spec/chore/queues/pubsub_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+  end
+
+  context "when managing queues" do
+    it 'should create topics and subscriptions that are defined in its internal job name list' do
+      topic_admin = double('topic_admin')
+      subscription_admin = double('subscription_admin')
+      allow(pubsub_client).to receive(:topic_admin).and_return(topic_admin)
+      allow(pubsub_client).to receive(:subscription_admin).and_return(subscription_admin)
+      allow(pubsub_client).to receive(:topic_path).with(queue_name).and_return("projects/test/topics/#{queue_name}")
+      allow(pubsub_client).to receive(:subscription_path).with(subscription_name).and_return("projects/test/subscriptions/#{subscription_name}")
```
obrie: There's a lot of repetition in these specs.  Would it be possible to clean it up a bit?

spec/chore/queues/pubsub_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+        subscription_admin = double('subscription_admin')
+        allow(pubsub_client).to receive(:topic_admin).and_return(topic_admin)
+        allow(pubsub_client).to receive(:subscription_admin).and_return(subscription_admin)
+        allow(pubsub_client).to receive(:topic_path).with(queue_name).and_return("projects/test/topics/#{queue_name}")
+        allow(pubsub_client).to receive(:subscription_path).with(subscription_name).and_return("projects/test/subscriptions/#{subscription_name}")
+        allow(topic_admin).to receive(:create_topic).and_raise(Google::Cloud::AlreadyExistsError.new('Topic already exists'))
+        allow(subscription_admin).to receive(:create_subscription).and_raise(Google::Cloud::AlreadyExistsError.new('Subscription already exists'))
+        expect(Chore.logger).to receive(:info).with("Chore Creating Pub/Sub Topic and Subscription: #{queue_name}")
+        expect(Chore.logger).to receive(:info).with("Topic already exists: Topic already exists")
+        expect(Chore.logger).to receive(:info).with("Subscription already exists: Subscription already exists")
```
obrie: These log expectations seem like they'll be brittle in the long-term.  I'm not sure if we get a ton of value out of having them.

lib/chore/queues/sqs.rb
```diff
@@ -1,7 +1,23 @@
...
+        # Verify compatible version
+        begin
+          gem_version = Gem::Version.new(Aws::SQS::GEM_VERSION)
+          if gem_version < MIN_VERSION
+            raise "#{REQUIRED_LIBRARY} version #{gem_version} is not supported. Please use version >= #{MIN_VERSION}"
+          end
+        rescue => e
+          Chore.logger.error "Could not verify #{REQUIRED_LIBRARY} version: #{e.message}" if defined?(Chore.logger)
+          exit
+        end
```
obrie, line 15: Do we need to raise and immediately rescue or can we do something like this?


```suggestion
          gem_version = Gem::Version.new(Aws::SQS::GEM_VERSION)
          if gem_version < MIN_VERSION
            Chore.logger.error "#{REQUIRED_LIBRARY} version #{gem_version} is not supported. Please use version >= #{MIN_VERSION}" if defined?(Chore.logger)
            exit
          end
```

The current logic seems more complicated that needed, but is there a use case we're trying to handle here?

lib/chore/queues/pubsub.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      # Creates a configured PubSub client with the given options
+      # Creates a GCP Pub/Sub client using global configuration
+      def self.pubsub_client
+        require REQUIRED_LIBRARY
+        
+        # Verify compatible version
+        begin
+          gem_version = Gem::Version.new(Google::Cloud::PubSub::VERSION)
+          if gem_version && gem_version < MIN_VERSION
```
obrie: Is there a case where `gem_version` is `nil`?  Can this jsut be `if gem_version < MIN_VERSION`?

lib/chore/queues/pubsub.rb
```diff
@@ -0,0 +1,132 @@
...
+      def self.existing_queues
+        client = pubsub_client
+        Chore.prefixed_queue_names.select do |queue_name|
+          begin
+            client.publisher(queue_name)
+            client.subscriber(queue_name)
+            # if both publisher/subscriber successfully load, then assume exists
+            true
+          rescue 
+            # google api throws Google::Cloud::NotFoundError if topic/subscription does not exist
```
obrie, line 120: Would it be possible to rescue specific errors here rather than `StandardError`?

lib/chore/queues/pubsub/publisher.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        # Retrieves the GCP Pub/Sub topic object. The method will cache the results to prevent round trips on subsequent calls
+        #
+        # If <tt>reset_connection!</tt> has been called, this will result in the connection being re-initialized,
+        # as well as clear any cached results from prior calls
+        #
+        # @param [String] name Name of GCP Pub/Sub topic
+        #
+        # @return [Google::Cloud::PubSub::Topic]
+        def get_topic(name)
```
obrie: I think this method name and the return type are no longer accurate

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,160 @@
...
+          unless subscription.exists?
+            raise "Subscription #{@subscription_name} does not exist"
+          end
+        end
+
+        # Begins requesting messages from GCP Pub/Sub, which will invoke the +&handler+ over each message
+        #
+        # @param [Block] &handler Message handler, used by the calling context (worker) to create & assigns a UnitOfWork
+        #
+        # @return [Array<Google::Cloud::PubSub::ReceivedMessage>]
```
obrie: This isn't accurate -- we don't return anything from this method

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,160 @@
...
+
+          messages
+        end
+
+        # Retrieves the GCP Pub/Sub subscription object. The method will cache the results to prevent round trips on subsequent calls
+        #
+        # If <tt>reset_connection!</tt> has been called, this will result in the connection being re-initialized,
+        # as well as clear any cached results from prior calls
+        #
+        # @return [Google::Cloud::PubSub::Subscription]
```
obrie: It looks like this returns `Google::Cloud::PubSub::Subscriber` as opposed to a `Subscription`

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,160 @@
...
+        def subscription
+          if !@pubsub_last_connected || (@@reset_at && @@reset_at >= @pubsub_last_connected)
+            @pubsub = nil
+            @pubsub_last_connected = Time.now
+            @subscription = nil
+          end
+
+          @subscription ||= pubsub.subscriber(@subscription_name)
```
obrie: I think this might be clearer if we call the instance variable a `subscriber` to match the type of object we're working with

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,160 @@
...
+
+        # Maximum number of messages to retrieve on each request from config
+        # Validates that the value doesn't exceed Pub/Sub's limit of 1000 messages
+        #
+        # @return [Integer]
+        # @raise [ArgumentError] if queue_polling_size exceeds 1000
+        def pubsub_polling_amount
+          polling_size = Chore.config.queue_polling_size
+          if polling_size > 1000
+            raise ArgumentError, "queue_polling_size (#{polling_size}) exceeds Google Cloud Pub/Sub maximum limit of 1000 messages"
```
obrie: I think we want to raise `Chore::TerribleMistake` here?  At least, that's what we tend to raise on unrecoverable issues

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,160 @@
...
+        end
+
+        # Maximum number of messages to retrieve on each request from config
+        # Validates that the value doesn't exceed Pub/Sub's limit of 1000 messages
+        #
+        # @return [Integer]
+        # @raise [ArgumentError] if queue_polling_size exceeds 1000
+        def pubsub_polling_amount
+          polling_size = Chore.config.queue_polling_size
+          if polling_size > 1000
```
obrie: Could you toss some of these magic numbers into constants defined at the top to make it easier to see/update these values?

lib/chore/queues/pubsub/consumer.rb
```diff
@@ -0,0 +1,160 @@
...
+        # If <tt>reset_connection!</tt> has been called, this will result in the connection being re-initialized,
+        # as well as clear any cached results from prior calls
+        #
+        # @return [Google::Cloud::PubSub::Subscription]
+        def subscription
+          if !@pubsub_last_connected || (@@reset_at && @@reset_at >= @pubsub_last_connected)
+            @pubsub = nil
+            @pubsub_last_connected = Time.now
+            @subscription = nil
+          end
```
obrie: Do you know if any of these is actually necessary for the pubsub sdk?  We had to do this with SQS because of how it manages connections, but I'm curious if the same requirements exists for pubsub.

lib/chore/queues/sqs.rb
ojameso: you're correct.

lib/chore/queues/pubsub/consumer.rb
ojameso: Hm, this was also on SQS so might have been copypasted over. Will remove there as well.

lib/chore/queues/pubsub/consumer.rb
ojameso: Good catch. From my understanding, pubsub handles conections and we don't have to worry about this.

lib/chore/queues/pubsub.rb
ojameso: Hm, I believe you're correct.

adnissen (commented):
lgtm

PR #74, opened 2025-08-18
[NPE-719] Add PayloadHandler

ojameso:
Adds `PayloadHandler` and makes it configurable so users can build custom payload handling for additional meta data or third party messaging consumption.

Chore is built upon the assumption that messages published to the queue will be handled by Chore. Even though Chore allows for configuration of `payload_handler`, it assumes a structure of the message that makes it difficult to do processing of messages that's not published by Chore (ie. third party streaming data). For instance, it assumes this `job_hash`:
```
{
  "class" => Klass,
  "args" => [arg1, arg2]
}
```
This makes it difficult to parse if the message does not correspond to this structure, for instance:
```
{
  "id" => "ID",
  "action" => "update",
  "params" => { "enabled" > true }
}
```
Chore determines the process to run by constantizing the class in the message which poses problems for messages published without this data. 

In order to facilitate the above, the method signature for decoding the message was updated to take the `UnitOfWork` with quite a bit of data. It also refactors code a bit to place message structure handling into `PayloadHandler` rather than `Job` to cleanly separate concerns. This way, a user has control over how messages should be constructed/deconstructed. The code still breaks if you don't follow the original `job_hash`, but it allows for great flexibility on the decoding of the message without reworking a ton of Chore internals. A custom payload handler may look like this.
```
module Chore
  class MyPayloadHandler < Chore::PayloadHandler
    # Override Chore's decoding to handle proper message handling for non-chore published messages
    def self.decode(item)
      klass = self.raw_json_queues[item.queue_name]
      if klass.present?
        args = Chore::Encoder::JsonEncoder.decode(item.message)
        # job_hash is defined on Chore::PayloadHandler and returns the expected Chore format for message
        return self.job_hash(klass, args) 
      else
        super
      end
    end

    def self.raw_json_queues
      # Class is stored as string for hot reloading in dev
      @raw_json_classes ||= {
        MyJob.prefixed_queue_name => "MyJob"
      }
    end
  end
end
```

### DIFF
--- a/lib/chore/worker.rb
+++ b/lib/chore/worker.rb
@@ -77,7 +77,7 @@ module Chore
       @work.each do |item|
         return if @stopping
         begin
-          item.decoded_message = options[:payload_handler].decode(item.message)
+          item.decoded_message = options[:payload_handler].decode(item)
           item.klass = options[:payload_handler].payload_class(item.decoded_message)
 
           next if duplicate_work?(item)
--- a/lib/chore.rb
+++ b/lib/chore.rb
@@ -14,6 +14,7 @@ require 'chore/publisher'
 require 'chore/util'
 require 'chore/worker'
 require 'chore/publisher'
+require 'chore/payload_handler'
 
 # We have a number of things that can live here. I don't want to track
 ['queues/**','strategies/**'].each do |p|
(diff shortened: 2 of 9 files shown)


lib/chore/worker.rb
```diff
@@ -77,8 +77,8 @@ def start
       @work.each do |item|
         return if @stopping
         begin
-          item.decoded_message = options[:payload_handler].decode(item.message)
-          item.klass = options[:payload_handler].payload_class(item.decoded_message)
+          item.job_details = options[:payload_handler].parse_job_details(item)
+          item.klass = options[:payload_handler].payload_class(item)
```
obrie: Why did we need to change the interface here?

obrie (commented):
Before I review further, can you expand in the PR description with some examples of what we're trying to support?  There are some interface changes being made here and it's not clear to me why they're needed.

lib/chore/worker.rb
ojameso: This was unnecessary and carryover from trying different things.

obrie (approved):

